PR #9, opened 2015-06-01
Explicitly declare babel-core and node-libs-browser in devDependencies

akovalyov:
```babel-core``` is required by ```babel-loader``` and its size is ~19 MB.
```node-libs-browser``` is required by ```webpack``` and its size is ~3MB.
If we explicitly define it as dev deps - it won't be copied to the release. As the result, the release will have ~22 MB less size, which is about 20%.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
   "devDependencies": {
     "asar": "^0.6.1",
     "babel-loader": "^5.1.2",
+    "babel-core": "^5.4.2",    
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by chentsulin** [comment]
It's great for size reduction. I will merge this after I add a section to `README.md` to avoid other people who using `babel-core` or `node-libs-browser` in runtime be confused. for example:

```js
require('bable/register')
```

PR #34, opened 2015-08-23
Placeholder app icon

danielbushman:
I created a logo to use as a placeholder icon for the app, and then made it the default icon for packaging in place of the Electron logo. This way I can have the hot-loading version running at the same time as a release and tell the difference between them. 

This really only includes a 512x512@2x icon. The smaller sizes are just downscaled. 

PNG example:
icon_512x512 2x


chentsulin:
It looks awesome.
Thanks for your contribution!

PR #41, opened 2015-09-10
use electron-prebuilt version for packaging. Fixes chentsulin/electro…

tsemerad:
This change just gets the current `electron-prebuilt` version by running `npm list | grep electron-prebuilt`. I looked for other ways to get the currently-installed version of a node package, and this is the only option I found. Tested and working as expected.


chentsulin:
Look great to me. Thanks!

PR #48, opened 2015-10-04
Simplify Electron executable scenario

H3Chief:
Created a bin link to `./node_modules/.bin/electron`, eliminating the need to have `electron` on your environment, and simultaneously making `scripts` easier to write.


chentsulin:
Fair enough, thanks!

Issue #57, opened 2015-10-17
Built Mac package shows yellow blank screen and doesn't load properly


RyanAtViceSoftware:
I followed the steps on the Readme and got the app to run in the dev mode but when trying to package and run it as an app I get a blank yellow screen with the errors shown below.

![image](https://cloud.githubusercontent.com/assets/10080111/10558647/6da51c96-749d-11e5-93ed-06ad7f4cc3fe.png)

I looked into the error and to get around this I had to switch the NODE_ENV to production so I changed the package.json like shown below.

**Before**
```
"package": "node package.js",
```

**After**
```
"package": "NODE_ENV=production node package.js",
```

Doing this got it to work but I'm not sure this is the best approach. Should I be setting the NODE_ENV in the terminal before I build? However, that feels wrong too.

chentsulin:
Oh, thanks for pointing out this. This should be a bug I introduced in 0.6.0..

Babel using `NODE_ENV` to determine whether should enable plugins or not :https://github.com/chentsulin/electron-react-boilerplate/blob/master/.babelrc#L4

`"package": "NODE_ENV=production node package.js"` looks great to me here.

RyanAtViceSoftware:
Oh cool, glad that's a viable solution. Did you want me to PR that?

chentsulin:
Sure, thanks for your help. Maybe we should take care `build`, `package` and `package-all` those three tasks.

RyanAtViceSoftware:
All the same way?

chentsulin:
The only one could be different is that `"package-all": "npm run package -- --all"` seems better.

chentsulin:
fixed via #59

PR #59, opened 2015-10-17
Set NODE_ENV to production for package, build targets

RyanAtViceSoftware:
@chentsulin I think this is what we discussed as a fix for #57. If not let me know and I can tweak it.


chentsulin:
Thanks!

PR #79, opened 2015-11-23
Upgrade Travis from legacy to container-based infrastructure

jhen0409:
refer: http://docs.travis-ci.com/user/migrating-from-legacy


chentsulin:
Thanks~ very helpful. I never use advanced travis config like this.

PR #87, opened 2015-12-05
remove unnecessary redundancy in App.js

iDuuck:
I thought about the "redundancy" in App.js (App.production and App.development are almost the same).
So I made this small pull request.

If the (in production) unnecessary `import` of the DevTools is not so cool for you, it would be possible that I also make a conditional in the render method with `const DevTools = require('./DevTools');`

What do you think?


chentsulin:
Putting `require` in a if statement can benefit from webpack dead code elimination, just like: https://github.com/gaearon/redux-devtools/blob/next/examples/counter/src/containers/Root.js

Though a little increase of size maybe doesn't matter here.

iDuuck:
But we could just do this in the single App.js, too!  
<br  
—

**Dominik Schmidt**  
Geschäftsführer

**  
**

**Kreativgebiet GmbH**  
Philosophenweg 31-33  
47051 Duisburg

  

Tel.: +49 172 305 386 0  
Web: kreativgebiet.com

> On Dec 6 2015, at 6:56 am, C. T. Lin &lt;notifications@github.com&gt; wrote:  

>

> Putting `require` in a if statement can benefit from webpack dead code
elimination, just like: <https://github.com/gaearon/redux-
devtools/blob/next/examples/counter/src/containers/Root.js>

>

> Though a little increase of size maybe doesn't matter here.

>

> —  
Reply to this email directly or [view it on
GitHub](https://github.com/chentsulin/electron-react-boilerplate/pull/87#issue
comment-162275759).

chentsulin:
Not sure one file will be more clear than two files here. @jhen0409 what do you think?

jhen0409:
I think we can merge it, because I cannot think reasons of often change `App.{env}.js`, but we need dead code elimination, it would be like the following code:
```js
render() {
  return (
    <div>
      {this.props.children}
      {
        (() => {
          if (process.env.NODE_ENV !== 'production') {
            const DevTools = require('./DevTools');
            return <DevTools />;
          }
        })()
      }
    </div>
  );
}
```

Also, for `react-redux-universal-hot-example` [example](https://github.com/erikras/react-redux-universal-hot-example/blob/master/src/client.js) it chose rerender component.

iDuuck:
Done!

chentsulin:
Thanks!

PR #93, opened 2015-12-08
use better-npm-run for better scripts management

### LINKED ISSUE #81: Error running npm run start-hot on Windows
C:\Users\x\devel\electron-react-boilerplate> npm run start-hot

> electron-react-boilerplate@0.7.0-beta start-hot C:\Users\x\devel\electron-react-boilerplate
> HOT=1 NODE_ENV=development electron .

'HOT' is not recognized as an internal or external command,
operable program or batch file.

npm ERR! Windows_NT 10.0.10240
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run" "start-hot"
npm ERR! node v5.0.0
npm ERR! npm  v3.3.6
npm ERR! code ELIFECYCLE
npm ERR! electron-react-boilerplate@0.7.0-beta start-hot: `HOT=1 NODE_ENV=development electron .`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the electron-react-boilerplate@0.7.0-beta start-hot script 'HOT=1 NODE_ENV=development electron .'.
npm ERR! This is most likely a problem with the electron-react-boilerplate package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     HOT=1 NODE_ENV=development electron .
npm ERR! You can get their info via:
npm ERR!     npm owner ls electron-react-boilerplate
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     C:\Users\x\devel\electron-react-boilerplate\npm-debug.log
PS C:\Users\x\devel\electron-react-boilerplate>

iDuuck:
May fix #81. Would be nice if someone could check this out.

### DIFF
--- a/package.js
+++ b/package.js
@@ -40,12 +40,13 @@ if (version) {
   startPack();
 } else {
   // use the same version as the currently-installed electron-prebuilt
-  exec('npm list | grep electron-prebuilt', (err, stdout, stderr) => {
+  exec('npm list electron-prebuilt', (err, stdout) => {
     if (err) {
       DEFAULT_OPTS.version = '0.35.2';
     } else {
-      DEFAULT_OPTS.version = stdout.split('@')[1].replace(/\s/g, '');
+      DEFAULT_OPTS.version = stdout.split('electron-prebuilt@')[1].replace(/\s/g, '');
     }
+
     startPack();
   });
 }
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
       "env": { "NODE_ENV": "production" }
     },
     "test": {
-      "command": "mocha --compilers js:babel/register --recursive --require ./test/setup.js $(find test -name '*.spec.js')",
+      "command": "mocha --compilers js:babel/register --recursive --require ./test/setup.js test/**/*.spec.js",
       "env": { "NODE_ENV": "test" }
     },
     "test-e2e": {
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "babel-eslint": "^4.1.1",
     "babel-loader": "^5.3.2",
     "babel-plugin-react-transform": "^1.1.1",
-    "better-npm-run": "0.0.4",
+    "better-npm-run": "0.0.5",
     "chai": "^3.3.0",
     "chromedriver": "^2.19.0",
     "co-mocha": "^1.1.2",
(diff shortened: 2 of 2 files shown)


jhen0409:
The problem isn't only related env, the `$(find test -name '*.spec.js')` isn't possible to run in Windows CMD.
Also, the `package.js` it used `npm list | grep electron-prebuilt`.

iDuuck:
And why aren't we just replacing the `mocha` command with the following?
```
mocha --compilers js:babel/register --recursive --require ./test/setup.js 'test/**/*.spec.js'
```

iDuuck:
Also, why is the pipe a problem for windows? I know, this document is for Windows XP, but I don't think the command line has changed *that* much since then: https://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/redirection.mspx?mfr=true

jhen0409:
@iDuuck not pipe, `grep` not work on Windows.

iDuuck:
what about utilizing something like https://www.npmjs.com/package/grep1 (quick search, there can be some better things out there)

iDuuck:
sorry, this is using the grep command using `child_process` so, this would not be a better solution

jhen0409:
Oh, we can use `test/**/*.spec.js`, but no quote char. (I tried it on Windows)

iDuuck:
And why aren't we just using the version number of the dependency from `package.json`. Is there any reason for this? (https://github.com/chentsulin/electron-react-boilerplate/blob/master/package.json#L59)

chentsulin:
See #41, "^" fixes the major number only.

iDuuck:
This should do it, right?

```
var pkg = require('./package.json');

exec('npm list electron-prebuilt', (err, stdout, stderr) => {
    if (err) {
      DEFAULT_OPTS.version = '0.35.2';
    } else {
      DEFAULT_OPTS.version = stdout.split('electron-prebuilt@')[1].replace(/\s/g, '');
    }
    
    startPack();
  });
```

chentsulin:
@iDuuck you are right. Could you also add those two changes into this PR?

jhen0409:
The `better-npm-run` has still problems in Windows. (https://github.com/benoror/better-npm-run/pull/24#issuecomment-163162679)

jhen0409:
Should work on `better-npm-run@0.0.5`, @iDuuck you can update it! :)

iDuuck:
Again: Done!

chentsulin:
Close via #100, Thanks~

PR #98, opened 2015-12-13
Add productName to package.json

jhen0409:
https://github.com/atom/electron/blob/master/docs/api/app.md#appgetname

Electron `app.getName()` default used productName of `package.json`, so we can use it to replace default appName of `package.js`.


Noitidart:
This isn't working for me. I changed my `package.json.productName` and `package.json.build.productName` but my `app.getPath('userData')` is still "Electron".

PR #113, opened 2015-12-28
Run chromedriver via node instead of command

jhen0409:


chentsulin:
:+1:

jhen0409:
Sorry, I ignored chromedriver start time (generally very fast), maybe we can delay `new webdriver.Builder`?

chentsulin:
Sure, we can try that. Just wander why test were passed on PR but failed on master

PR #114, opened 2015-12-28
Delay webdriver build (wait chromedriver start time)

jhen0409:
related https://github.com/chentsulin/electron-react-boilerplate/pull/113#issuecomment-167597333


jhen0409:
The chromedriver start by child_process.execFile, and we cannot check it started listening 9515 port, sometimes it probably error on Travis, so I delay webdriver build. (just like old .travis.yml)

PR #121, opened 2016-01-09
package script now chooses correct OS icon extension

Kilian:
The packaging script now appends a different icon extension depending on the platform:

* win32 needs an .ico
* darwin needs an .icns
* everything else (i.e., linux) needs a png.

This commit assumes an extensionless icon argument, and 3 icon files on disk (like the current script assumes 1) if you're doing a package-all and chooses the correct extension for the platform it's currently being built for.

### DIFF
--- a/package.js
+++ b/package.js
@@ -92,7 +92,7 @@ function pack(plat, arch, cb) {
         extension = '.ico';
       }
       return extension;
-    }())
+    })()
   };
   
   const opts = Object.assign({}, DEFAULT_OPTS, iconObj, {
--- a/package.js
+++ b/package.js
@@ -92,7 +92,7 @@ function pack(plat, arch, cb) {
         extension = '.ico';
       }
       return extension;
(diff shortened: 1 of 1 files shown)


package.js
```diff
@@ -83,7 +83,19 @@ function pack(plat, arch, cb) {
...
+  const iconObj = {
+    icon: DEFAULT_OPTS.icon + (() => {
+      let extension = '.png';
+      if (plat === 'darwin') {
+        extension = '.icns';
+      } else if (plat === 'win32') {
+        extension = '.ico';
+      }
+      return extension;
+    })
```
jhen0409: It should be `(() => ...)()`?

Kilian: You're right, sorry!

jhen0409:
[have one lint error](https://travis-ci.org/chentsulin/electron-react-boilerplate/jobs/101387956#L250)

PR #122, opened 2016-01-12
Use webpack-dev-server CLI instead of express server

jhen0409:
Also, the `hot-dev-app.html` it seems pointless in `localhost:3000`, or @chentsulin do you has other considerations?


chentsulin:
Just want to follow what gaearon does in react-transform-boilerplate..
https://github.com/gaearon/react-transform-boilerplate/blob/master/devServer.js

davej:
This is nice, thanks @jhen0409 :+1:

davej:
Quick note on this, you can remove `new webpack.HotModuleReplacementPlugin()` from plugins. In fact if you don't then you will probably encounter call stack size errors.

chentsulin:
IMO, following up with redux official examples and react-transform-boilerplate is better here.

davej:
Cool. Just as an FYI:

@gaearon has said that the main reason he uses an Express server in the boilerplate is because people often have a Node server anyway, and it can be tricky to configure webpack-dev-server to work with an existing node server. So, perhaps that reason isn't really relevant to an Electron app. Anyway, both approaches work fine so it's a bit of a bike-shed issue really. :-)

PR #124, opened 2016-01-14
Fix Webpack publicPath for production builds

pascalw:
Previous value `/dist/` would cause static assets to be searched for in
`file:///dist/...` (root of fs), instead of in the `dist/` directory
relative to `app/index.html`.


etisdew:
Necropost but I bout racked myself over this silly little bug. Thanks for the the use of issues @pascalw and context to my madness.

PR #126, opened 2016-01-15
react-router 2

chentsulin:


app/index.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { render } from 'react-dom';
 import { Provider } from 'react-redux';
 import { Router } from 'react-router';
+import { hashHistory } from 'react-router';
```
borNfreee: Just curious, why hashHistory?
https://github.com/rackt/react-router/blob/latest/docs/guides/basics/Histories.md#should-i-use-createhashhistory
> Hash history is the default because it works without any setup on your server, and works in all evergreen browsers and IE8+. But, we don't recommend using it in production, every web app should aspire to use createBrowserHistory.

chentsulin: @borNfreee You can't use `browserHistory` in electron apps. Because routes matching will rely on file path.

```
Warning: [react-router] Location "/Users/xxx/electron-react-boilerplate/app/hot-dev-app.html" did not match any routes
```

borNfreee: thank you for the explanation @chentsulin

slightlytyler:
@chentsulin any interest in using (react-router-redux)[https://github.com/rackt/react-router-redux]? Keeping routing state in the store is very useful imo (: I don't mind adding it

chentsulin:
@slightlytyler feel free to open a new PR. I will merge it.

app/index.js
jbovenschen: https://github.com/reactjs/react-router/blob/master/docs/guides/Histories.md#creatememoryhistory isn't this one better, because you don't heave a real history in the electron environment, if i am not mistaken?

PR #135, opened 2016-02-04
Use babel-plugin-webpack-loaders instead of css-modules-require-hook

jhen0409:
It will can be support more loaders (e.g. url-loader) on test.


chentsulin:
:+1:

PR #142, opened 2016-02-13
Use babel-plugin-transform-runtime instead of babel-polyfill

jhen0409:
Related https://github.com/SeleniumHQ/selenium/issues/1627.

Maybe caused by a recent babel update, the E2E test is not work.

If someone needs `core-js`, I would recommend use v2.0, or wait babel update.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": ["es2015", "stage-0", "react"],
-  "plugins": ["add-module-exports"],
+  "plugins": ["add-module-exports", "transform-runtime"],
   "env": {
     "development": {
       "presets": ["react-hmre"]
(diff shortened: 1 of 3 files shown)


chentsulin:
Could we just wait for the upcoming release version?
https://github.com/babel/babel/commit/79214b3b05f1c46c6c1176f1c1d05080c284849e#diff-add20b3ca7c5515bbe589bc5ea49bb6b

chentsulin:
@jhen0409 will v6.6.1 solve this? https://github.com/babel/babel/blob/master/packages/babel-polyfill/package.json

jhen0409:
It might be solved, but another problem occur (https://github.com/istarkov/babel-plugin-webpack-loaders/issues/39), so I still cannot test it.

chentsulin:
Should be fixed via v6.6.3? @jhen0409 
https://phabricator.babeljs.io/T7164

jhen0409:
@chentsulin yes. :)

PR #154, opened 2016-02-22
Set style.global.css rather than style.module.css

epilande:
Just an idea.
Add `.global` for global styles.
Anything without `.global` will go through css-modules.
Basically reversing the current css implementation.

Projects will most likely have more css modules than global styles. This way we wont have to add `.module` to all of our css.


chentsulin:
LGTM. This one make more sense and encourage people avoid global css. @jhen0409 What do you think?

If we want to go this way, we must also update README's css module section.

jhen0409:
:+1: Great idea! I don't even often write global css, and it's more convenient to name each component.

epilande:
Updated the README :smiley:

PR #161, opened 2016-02-26
Use dynamic script switches instead of`hot-dev-app.html`. Closes #160

### LINKED ISSUE #160: Use dynamic script switch instead of `hot-dev-app.html`
Instead of having a separate html file for dev purposes you could do something like this in `app.html`:

```html
<script>
  const script = document.createElement('script');
  script.src = (process.env.HOT) ? 'http://localhost:3000/dist/app.js' : '../dist/app.js';
  document.write(script.outerHTML);
</script>
```

davej:
* I applied the same technique to the css file also
* I used an IIFE to prevent the `link` and `script` variables from leaking to the outer scope (window).


davej:
Woops, failing tests. I'll look into it now.

PR #162, opened 2016-02-27
Added start (hot-server, start-hot) servers with one command

suribit:
Keep open two terminals is bad.
So I added the ability to run two servers with a single command (npm run dev). 
I hope this opportunity will be useful to others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,8 @@ $ npm install
 
 ## Run
 
-Run this two commands __simultaneously__ in different console tabs.
-
 ```bash
-$ npm run hot-server
-$ npm run start-hot
+$ npm run dev
 ```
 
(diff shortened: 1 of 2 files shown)


chentsulin:
Why `better-npm-run` npm module is needed here?

suribit:
Yes you are right. `better-npm-run` is not necessary here, because It has `cross-env`.

jhen0409:
In my use case, I often change main process, and restart it, but webpack server not need. If they are restart together, there will be a longer waiting time.

suribit:
@jhen0409 yes, the restart time will increase this way, but it's just one option run. You can also run separately (`npm run hot-server` and `npm run start-hot`). 
I have made changes in README.MD on this occasion

PR #170, opened 2016-03-03
Remove unused handler

Yeti-or:
Hello I don't understand why we need this handler
`app.use(require('webpack-hot-middleware')(compiler));` isn't enough?


chentsulin:
You are right. Merged!

Issue #119, opened 2015-12-31
Webpack publicPath configured incorrectly for production builds?


pascalw:
I'm new to Webpack so I could be wrong but; I'm including font-awesome in my project and noticed that in production builds the icons would not work, because it tried to load the icon files from `/dist/...ttf`.

Looking at the `webpack.config.production.js` I noticed this snippet which I think is incorrect:

`config.output.publicPath = '/dist/';`

Setting that to `../dist/` fixes my problem with font-awesome. Looking at the [Webpack docs](https://github.com/webpack/docs/wiki/configuration) I think this indeed should be `../dist/` because we're not running a server in production builds.

Do you want me to send a PR for that?

chentsulin:
This is a non-issue, and just about directory structure preference. I prefer to put all static resources into a `dist` directory instead of everywhere. So, personally I will copy font files into dist directory, too.

Notice that all js and css files will bundle into `dist` via webpack in this boilerplate, if you want to change publicPath, you should be careful and correctly handle path and directory change. 

Another thing worth to mention here is that `../dist/` have no different with './'.

pascalw:
I think I may have not been clear enough.

The problem is that `config.output.publicPath` for production builds is [configured to point to `/dist/`](https://github.com/chentsulin/electron-react-boilerplate/blob/2989a02bd5703c7896b61be8aa5c55248e705bdb/webpack.config.production.js#L16) (notice the leading slash). However this doesn't work because in production builds assets are served from the filesystem, not a webserver.

jhen0409:
@pascalw a real electron application doesn't need webserver.

pascalw:
@jhen0409 yes exactly and that's why `config.output.publicPath = '/dist/'` doesn't work for production builds, as production builds run in the context of a file system meaning `/dist/` would refer to the root of the filesystem.

L-A:
Just chiming in to say that this should probably count as an issue – everyone using assets will bump into this and need to change the path, even if everything is in `dist`.

Since packaged assets are not requested from a server, files that are compiled to `dist` will currently be requested from `file:///dist/assetname`, because it's configured as an absolute path.

As @pascalw is saying, with `../dist/` in `config.output.publicPath`, this would behave as expected.

pascalw:
@chentsulin send a [PR](https://github.com/chentsulin/electron-react-boilerplate/pull/124) that fixes this.

L-A:
[...]

PR #173, opened 2016-03-04
Remove output.publicPath from production config

Yeti-or:
As we have `output.path` in base.config


jhen0409:
It's necessary, see the discussion of #119.

Yeti-or:
I see now, thanx @jhen0409

PR #186, opened 2016-03-23
Fixed using the same version as the currently-installed electron-prebuilt

IGZangelsanchez:


chentsulin:
Note that `^0.37.2` may install `0.37.3` when a patch version has been released, so this PR can't ensure it package  electron with same version as electron-prebuilt.

IGZangelsanchez:
You are right. But the actual implementation throws an error in Ubuntu. After execution of `npm list electron-prebuilt` so the version set is always the hardcoded version `DEFAULT_OPTS.version = '0.37.2';`. The error is:
```
{ [Error: Command failed: /bin/sh -c npm list electron-prebuilt
npm ERR! code 1
]
  killed: false,
  code: 1,
  signal: null,
  cmd: '/bin/sh -c npm list electron-prebuilt' } electron-react-boilerplate@0.9.0 /home/asanchez/igz/projects/electron-react-boilerplate
└── (empty)
```

jhen0409:
> After execution of npm list electron-prebuilt so the version set is always the hardcoded version DEFAULT_OPTS.version = '0.37.2';. The error is:

This is right, looks the `npm run package` is included `npm config dev: ""` env. (if I just use `node package.js`, it work)
Use `npm list electron-prebuilt --dev` it work.

PR #188, opened 2016-03-25
Fix `npm list electron-prebuilt` in package.js

jhen0409:
The `npm run package` is included `npm_config_dev: ""` env, it means it don't check devDependencies.

Related to https://github.com/chentsulin/electron-react-boilerplate/pull/186#issuecomment-201340977.


chentsulin:
Great :+1:

PR #192, opened 2016-03-30
Open application when webcontent is loaded

amine7536:
Open mainWindow when webContents is loaded using the event 'did-finish-load'.
This prevent having a blank window and then content "flash" in.


chentsulin:
This makes a lot of sense. Thanks! :+1:

PR #195, opened 2016-04-07
use es2015 syntax for webpack configs

domasx2:
Hi!

I think webpack configs are nicer to manage and read if spread operator is used to extend them. 

```javascript
//instead of
const config = Object.create(baseConfig);
config.entry = ['another_entrypoint.js'];
config.plugins.push(new SomePlugin()};
module.exports = config;

//use this
export default {
   ...baseConfig,
   entry: ['another_entrypoint.js'],
   plugins: [...baseConfig.plugins, new SomePlugin()]
};
```
This is easy to add too since babel is already in the project.

This PR:

*  babel-node used to run webpack, instead of node
* webpack configs rewritten using es2015/es2017 syntax


chentsulin:
Thanks!

Issue #139, opened 2016-02-07
Use Babel for main process


davej:
What's the consensus about letting the main process run through babel too? The issue I'm having is that I want to use `redux-electron-store` but I can't require any of the reducers because they use the ES6 module syntax which currently isn't supported by the latest versions of Electron/Chromium.

jhen0409:
Yes,  if we don‘t use Babel, this is no way to share reducers between Main and Renderer.

@chentsulin what do you think?

chentsulin:
I don't mind using babel also in main process, but which one approach is our best choice? `babel-register`?

davej:
The problem with using babel-register is that babel then needs to be bundled with the app (rather than simply as a dev-dependency).

davej:
Actually, I guess we could use `babel-register` in development and then compile it (using webpack?) for the production version.

wasd171:
Code sharing is quite essential. Unfortunately, simple adding of `require('babel-register');` in `main.js` results in 
```
App threw an error when running TypeError: Path must be a string. Received undefined
```
requiring the following `dummy.js`:
```
module.exports = 'bar';
```
I've tried webpack'ing `main.js`, but no luck :(

naderhen:
Is there a known workaround for this? @davej your solution seems reasonable, but I'm curious how you'd go about it within this boilerplate.

davej:
@naderhen: it just worked for me.

My entry point file looks like this:

```js
/* eslint strict: 0 */
'use strict';

require('babel-register');
require('./main/init.js');
```

`./main/init.js` is where my `main` process logic is.

naderhen:
I have something very similar which does indeed work in terms of running the electron application with babel included. However, when I attempt to package the app via `npm run package` (on OSX) and run the resulting *.app I get a host of babel related import errors:

![image](https://cloud.githubusercontent.com/assets/99392/14397101/80c55170-fda9-11e5-807c-0f27c36c539d.png)

Any thoughts? Do you run into the same issue when packaging?

davej:
@naderhen: I haven't tried packaging the app yet. My plan was to use webpack and create an entry for the main js file. Have you tried using webpack to compile the main process files instead of including babel in your production build?

naderhen:
[...]

PR #197, opened 2016-04-09
Use Babel in main process with Webpack build

jhen0409:
Related to #139, and just use `-r babel-core/register` in dev mode.

cc @chentsulin @davej


webpack.config.electron.js
```diff
@@ -0,0 +1,31 @@
...
+    { raw: true, entryOnly: false }
+  ),
+  new webpack.DefinePlugin({
+    'process.env': {
+      NODE_ENV: JSON.stringify('production')
+    }
+  })
+);
+config.externals.push('electron');
+config.target = 'node';
```
chentsulin: should use 'electron' as target

jhen0409: fixed :)

jhen0409:
Also, maybe we can just pack externals of webpack config on `npm run package`?

webpack.config.production.js
```diff
@@ -15,7 +15,7 @@ config.entry = './app/index';
 
 config.output.publicPath = '../dist/';
 
-config.module.loaders.push({
+config.module.loaders = config.module.loaders.concat([{
```
jhen0409: Looks `.production.js` and `.electron.js` will have problem of object reference in `package.js`, so I handle these with `array.concat`.

chentsulin:
I just wonder whether `babel-regiseter` is good enough for development..If we use some webpack specified feature in production, we can't directly apply them to development env.

package.json
```diff
@@ -10,9 +10,11 @@
     "test-e2e": "cross-env NODE_ENV=test mocha --compilers js:babel-core/register --require ./test/setup.js --require co-mocha ./test/e2e.js",
     "lint": "eslint app test *.js",
     "hot-server": "babel-node server.js",
-    "build": "cross-env NODE_ENV=production babel-node ./node_modules/.bin/webpack --config webpack.config.production.js --progress --profile --colors",
+    "build-electron": "cross-env NODE_ENV=production babel-node ./node_modules/.bin/webpack --config webpack.config.electron.js --progress --profile --colors",
```
chentsulin: Are `build-main` and `build-renderer` nicer?

jhen0409:
> If we use some webpack specified feature in production, we can't directly apply them to development env.

Yes, I can think of case is webpack loader, but it can be resolved by `babal-plugin-webpack-loaders`.

In addition, I recommend webpack build of main process is for production. @chentsulin What do you think?

chentsulin:
I'm ok for that. I don't know why but there is a regression about #192. It seems window will open before content ready.

jhen0409:
> It seems window will open before content ready.

It's `ERR_CONNECTION_REFUSED` error? It happens sometimes.

The `npm run hot-server` used `babel-node` (#195), so it's time to start may later than main process.

chentsulin:
Could we somehow fix `ERR_CONNECTION_REFUSED` to improve user experience here?

jhen0409:
Use `node -r babel-register` will be faster, it just not has some options and require `babel-polyfill`.

package.json
```diff
@@ -8,7 +8,7 @@
     "test": "cross-env NODE_ENV=test mocha --compilers js:babel-register --recursive --require ./test/setup.js test/**/*.spec.js",
     "test-watch": "npm test -- --watch",
     "test-e2e": "cross-env NODE_ENV=test mocha --compilers js:babel-register --require ./test/setup.js --require co-mocha ./test/e2e.js",
-    "lint": "eslint app test *.js",
+    "lint": "eslint app test *.js --ignore-pattern main.js",
```
chentsulin: Why not put this in a `.eslintignore`?

chentsulin:
I am almost ready to merge this PR. @jhen0409 Could you move `--ignore-pattern` to a `.eslintignore`? Thanks a lot.

PR #206, opened 2016-04-19
Fix ignore regex to top level folders

crucialfelix:
This was ignoring all folders named `release` in all node_modules

For instance bluebird ships with `js/browser` and `js/release` but `js/release` was ignored during the build process and filtered out.

https://github.com/petkaantonov/bluebird/blob/master/package.json#L71


chentsulin:
Thanks!

PR #216, opened 2016-05-10
Upgrade to react-hot-loader 3 from react-transform

tyscorp:


jhen0409:
We should not be required `react-hot-loader` in the production, maybe we can separate to `.prod.js` and `.dev.js`, or just a `NODE_ENV` check?

tyscorp:
I think that AppContainer and the entry patch amount to a noop when NODE_ENV is production by default.

jhen0409:
@tyscorp oh, indeed. lol

But it will still saved [some code](https://github.com/gaearon/react-hot-loader/blob/next/src/index.js) for not `require('react-hot-loader')`, though not many.

chentsulin:
We will revisit this PR after react-hot-loader v3 release!

madc:
@tyscorp I've implemented this functionality based on this PR in my project and it works in development env. But when running in production (`npm run build` && `npm run start`) I get the following:

```bundle.js:1 Uncaught ReferenceError: AppContainer is not defined```

Does this work for you?

amilajack:
What is the difference between react-hot-loader 3 and react-transform?

amilajack:
@tyscorp can you please rebase this PR

amilajack:
Closing, as `react-transform` has been deprecated

PR #220, opened 2016-05-13
Upgraded electron to v1 and remove crashReporter

chentsulin:
Removed crashReporter (we don't want to setup a server here to keep the report)
https://github.com/sindresorhus/generator-electron/issues/19


xdvarpunen:
Nothing mentioned about target: 'electron-renderer' here...

PR #221, opened 2016-05-15
Fix not stopping dev server on sigterm.

Yeti-or:
If you run `npm run dev`, and then close app with `cmd+q`, dev server is still opened.

### DIFF
--- a/server.js
+++ b/server.js
@@ -11,12 +11,14 @@ const app = express();
 const compiler = webpack(config);
 const PORT = 3000;
 
-app.use(webpackDevMiddleware(compiler, {
+const wdm = webpackDevMiddleware(compiler, {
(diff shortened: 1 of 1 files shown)


chentsulin:
We already use `concurrently --kill-others` to run the `dev` script. Could this case still happened?

Yeti-or:
Yes `concurrently` sends `SIGTERM` to our dev server but It doesn't close well.
Check this video: https://yadi.sk/d/Xkc3qCZ1rjNk8
also I added closing webpackDevMiddleware. Without it build after quiting app does not wait until bundle, in next start. You can see it on second `npm run dev` in video

chentsulin:
Thanks!

PR #227, opened 2016-05-26
Add devtron and electron-react-devtools

davej:
A couple of notes with devtron:

* The require graph doesn't really work on the renderer because electron-react-boilerplate uses a `script` tag instead of the `require` syntax for it's entry point.
* Loading event listeners aren't working for some reason, I've created an issue here: https://github.com/electron/devtron/issues/73


davej:
Not sure why the Node v4 build is failing in Travis.

Probably better to remove that build anyway since Electron hasn't been using Node 4 since v0.35 (last year).

davej:
Did some more testing just now. This actually doesn't work for the packaged version of the app because `process.env.NODE_ENV` isn't defined in html files. Any suggetions about where the code could go instead?

Keep in mind that `electronReactDevTools.inject()` needs to be called before React.

davej:
I've just seen that 1.0 of [`electron-debug`](https://github.com/sindresorhus/electron-debug) also activates devtron for you. Perhaps this is a better way to do it.

Unfortunately I don't think there's a way to install `electronReactDevTools` in just the main process but perhaps it would be possible by using something like `webContents.executeJavaScript(…)`.

davej:
Looks like React Devtools will work out of the box with the next version of Electron: https://github.com/electron/electron/pull/5711

I'll close this and may create a new PR when the new version of Electron is released.

chentsulin:
@davej thanks for looking into this. It's very helpful.

chentsulin:
@davej want to submit a PR for this? electron v1.2.1 has been released.

davej:
Sure, I'm on holidays at the moment so it will be Wednesday before I can do it though.

davej:
Just to follow up on this. The problem with doing this in app code is that it prints the following to console (after being run more than once):

```
Attempted to load extension "devtron", but extension was already loaded!
Attempted to load extension "React Developer Tools", but extension was already loaded!
```

Everything works fine but it's a bit messy. I've created an upstream issue in Electron so we can do this a bit more gracefully: https://github.com/electron/electron/issues/5960

I'll reopen this PR so it can be tracked.

davej:
@chentsulin This is ready for review now. [`electron-debug`](https://github.com/sindresorhus/electron-debug) will add `devtron` so I didn't duplicate that functionality. Also see: https://github.com/sindresorhus/electron-debug/pull/40

Just wondering, is there a reason why `electron-debug` is in deps and not devDeps?

jhen0409:
@davej Consider whether to install the official react-devtools via [electron-devtools-installer] (https://github.com/GPMDP/electron-devtools-installer)?

davej:
@jhen0409 Just tried it. It needs to be added to externals to work because the `cross-unzip` submodule isn't compatible with webpack. Also the react devtools tab doesn't appear until the second time the app is launched.

What do you think?

jhen0409:
Add `.json` to `resolve.extensions` on webpack config should work (because [this](https://github.com/fritx/win-7zip/blob/dev/index.js#L2)). Also I opened https://github.com/GPMDP/electron-devtools-installer/issues/1, it should be the same problem. :)

davej:
@jhen0409 Ok thanks, it works with the new version of electron-devtools-installer.

The only problem is that it doesn't seem to add the devtool on first run, you have to run the app twice. From a quick read of the electron-devtools-installer docs this doesn't seem to be mentioned I wonder is it a bug somewhere?

davej:
As an aside, I wonder is worth adding this code to a separate js file and requiring it instead of clogging up the html file.

jhen0409:
It seems the background script of extension cannot know new page onload, it's installed after the page load, maybe we can reload the page? Or move installation to main process (need change electron-devtools-installer).

Also, the electron-react-devtools [call `inject` after installation](https://github.com/firejune/electron-react-devtools/blob/master/api.js#L19), so it works.

davej:
Thanks @jhen0409!

Ok, how about this:

```js
if (process.env.NODE_ENV === 'development') {
  const getDevToolsExtensions = require('electron').remote.BrowserWindow.getDevToolsExtensions;
  const electronDevtoolsInstaller = require('electron-devtools-installer');
  
  const reactDevtoolsAlreadyAdded =
    getDevToolsExtensions && getDevToolsExtensions().hasOwnProperty('React Developer Tools');
  
  if (!reactDevtoolsAlreadyAdded) {
    electronDevtoolsInstaller(electronDevtoolsInstaller.REACT_DEVELOPER_TOOLS);
    if (getDevToolsExtensions) location.reload();
  }
}
```

The thing to be wary of is that on Electron 1.2.0 - 1.2.2 we can't do a reload because it will go into an infinite loop of reloading because the `getDevToolsExtensions` API isn't available. So I put in a check for that first. On Electron 1.2.0 - 1.2.2 the user will still need to manually reload the page.

I also tried installing from the main process but it didn't work for me.

MarshallOfSound:
@davej I just published a new version of `electron-devtools-installer` that handles not installing the same extension twice.  The current behavior is too reject the promise if the extension is already installed though I'm not sure that is the desired behavior (open to feedback)

The devtools not appearing on first load is an electron bug being addressed here https://github.com/electron/electron/pull/6115

davej:
@MarshallOfSound: Ah right cool, so that means the best approach is probably this:

```js
if (process.env.NODE_ENV === 'development') {
  const electronDevtoolsInstaller = require('electron-devtools-installer');
  electronDevtoolsInstaller(electronDevtoolsInstaller.REACT_DEVELOPER_TOOLS).catch(() => {});
}
```

For now users will have to manually reload but it looks like that will be fixed in the next version of electron.

Is this correct?

MarshallOfSound:
Yeah, that sounds about right.

Thought the module is designed to be used with the ES6 `import` style of things.  So you may have to do something silly like.

```js
if (process.env.NODE_ENV === 'development') {
  const devtoolsInstaller = require('electron-devtools-installer');
  devtoolsInstaller.default(devtoolsInstaller.REACT_DEVELOPER_TOOLS).catch(() => {});
}
```

davej:
Ok I think this is finally ready to go, thanks @jhen0409 and @MarshallOfSound. A lot of conversation for a couple of lines of code! 😀

`electron-debug` also now checks if the extension is already installed so there will be no console clutter.

jhen0409:
@davej @MarshallOfSound nice work! 👍

PR #233, opened 2016-05-30
extend output config in electron webpack config

the4dpatrick:
This is a fix for an issue I was running into with issue https://github.com/chentsulin/electron-react-boilerplate/issues/232

It adds an `output.libraryTarget` setting into `webpack.config.electron.js` by extending `baseConfig`

### DIFF
--- a/webpack.config.electron.js
+++ b/webpack.config.electron.js
@@ -9,6 +9,7 @@ export default {
   entry: './main.development',
 
   output: {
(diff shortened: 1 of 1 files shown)


davej:
FYI: I also ran into this issue (external modules with a 'dash' in the package name) but forgot to create a PR for it. So 👍 from me.

chentsulin:
Thanks 👍

PR #247, opened 2016-06-15
Added babel plugin react-optimize for production env

amilajack:


jhen0409:
hmm... [this is already on here](https://github.com/chentsulin/electron-react-boilerplate/blob/master/.babelrc#L6).

amilajack:
Oh sorry about that. Somehow my branch was out of date.

PR #248, opened 2016-06-17
Added why-did-you-update, updated deps

amilajack:


jhen0409:
Maybe make it optional will be better? Because it will has very long logs on dev mode.
(e.g. `$ INSPECT_REACT_UPDATE=1 npm run start-hot`)

chentsulin:
Personally, I want to avoid too much complexity increment in this project, so I will only accept this as a REAME section. Thanks anyway.

PR #263, opened 2016-07-01
Add UPGRADE_EXTENSIONS env for devtools-installer

jhen0409:
Use `UPGRADE_EXTENSIONS=1 npm run start-hot` to force download extensions.


jhen0409:
electron-devtools-installer@1.1.5 is out, merging

Issue #217, opened 2016-05-11
Electron 1.0


chentsulin:
http://electron.atom.io/blog/2016/05/11/electron-1-0

Maybe we can give those a try:
- spectron: https://github.com/electron/spectron
- devtron: https://github.com/electron/devtron

stu-nguyen:
Hi @chentsulin , 

Is that correct if we just simply need to update the version of electron dependancies?

chentsulin:
@stu-nguyen see #220, simply update the version and remove crashReporter, then, everything will work.

You can find Migration notice here: https://github.com/electron/electron/releases/tag/v1.0.0 

Please update your app to v0.37.8 first to check whether there are deprecated APIs usages.

Following deprecated events of BrowserWindow have been removed but there were no deprecation warnings in previous versions, you should use the corresponding events in webContents instead:
- crash
- devtools-focused
- devtools-opened
- devtools-closed

NativeImage.toDataUrl has been removed but there were no deprecation warnings in previous versions, you should use toDataURL instead.

stu-nguyen:
@chentsulin 👍 Thank you!!! I'll check it out.

xdvarpunen:
target:'electron-renderer' not mentioned...

amilajack:
Closing, as we have already migrated to electron 1+

PR #265, opened 2016-07-04
Use spectron instead of selenium-webdriver

jhen0409:
Related to #217.

It's mean we will migrate to WebdriverIO for writing E2E tests.


chentsulin:
🎉 🎉  🎉 🎉  🎉 🎉

PR #280, opened 2016-07-15
Added appveyor support

amilajack:


jhen0409:
👍

We can consider add `os: unstable` option it's recommend by [spectron](https://github.com/electron/spectron#on-appveyor).

/cc @chentsulin

amilajack:
Fixed

chentsulin:
Maybe we can cache `node_modules` folder as what we does with travis?

https://github.com/chentsulin/electron-react-boilerplate/blob/master/.travis.yml#L10

amilajack:
Added caching support

amilajack:
Also remember to enable the tests at https://www.appveyor.com

PR #297, opened 2016-07-21
Added 'compact' mode for faster buidls

amilajack:
https://blog.mariusschulz.com/2016/07/12/speeding-up-babel-transpilation-with-compact-mode

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,6 @@
 {
   "presets": ["es2015", "stage-0", "react"],
+  "compact": true,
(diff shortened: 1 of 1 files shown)


davej:
From the mentioned blog post:

> Update (July 21, 2016): It looks like there was an issue with Babel's code generation taking exponential time, which was fixed recently. After updating to a new Babel version, I no longer see the performance benefits described below.

amilajack:
I see. Closing PR.

PR #299, opened 2016-07-21
Fix .eslintrc import/resolver settings

akozhemiakin:
Eslint webpack import resolver looks for 'webpack.config.js' file as a sibling of the first ancestral package.json. We have no 'webpack.config.js' here so we need to point resolver to some
custom path. Also, we can't point it directly to the 'webpack.config.development.js' file cause it uses ES6 stuff. In order to make eslint import resolver happy the 'webpack.config.eslint.js' is added as a babel-register wrapper around 'webpack.config.development.js'.

Also, invalid json (missing or single quotes) is fixed here as a bonus.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,7 +10,7 @@
     "consistent-return": 0,
     "comma-dangle": 0,
     "no-use-before-define": 0,
-    "import/no-unresolved": [2, { ignore: ['electron'] }],
+    "import/no-unresolved": [2, { "ignore": ["electron"] }],
     "react/jsx-no-bind": 0,
     "react/prefer-stateless-function": 0
   },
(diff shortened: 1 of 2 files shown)


amilajack:
Is it possible to use es6 imports instead?:

```
Import 'babel-register';
Import config from  './webpack.config.development'
export default config;
```

akozhemiakin:
No. Actually the only reason of creating separate config file for eslint is that eslint doesn't use babel or any other transpiler to be able to deal with es6 stuff (import\export statements in particular).

And `require('babel-register');` in `webpack.config.eslint.js` allows us to enable transpiler on the fly require es6 modules like `./webpack.config.development`.

chentsulin:
@akozhemiakin Nice job 👍

PR #309, opened 2016-07-27
Use electron-builder to build app

xwartz:
Closed https://github.com/chentsulin/electron-react-boilerplate/issues/209


xwartz:
I have no idea about e2e test. If anyone can help?

dustintownsend:
@xwartz this is great! I made some changes here:
https://github.com/dustintownsend/electron-react-boilerplate/tree/PR309 

- the static/node_modules folder shouldn't be committed. Removed and updated .gitignore
- postinstall script needs to call install-app-deps to install the node_modules in static
- changed devDependenices to dependencies in static/package.json so the install-app-deps script would work
- fixed the e2e test. Need to update line 15 to pass in 'static' instead of '.' for the args.

PR #310, opened 2016-07-27
Use electron-builder to build app

dustintownsend:
Made some changes to @xwartz pull request #309 

I apologize if I did this incorrectly. This is my first time submitting a pull request.


xwartz:
Thanks, It's great!

package.json
```diff
@@ -14,11 +14,51 @@
     "build-renderer": "cross-env NODE_ENV=production node -r babel-register ./node_modules/webpack/bin/webpack --config webpack.config.production.js --progress --profile --colors",
     "build": "npm run build-main && npm run build-renderer",
     "start": "cross-env NODE_ENV=production electron ./",
-    "start-hot": "cross-env HOT=1 NODE_ENV=development electron -r babel-register -r babel-polyfill ./main.development",
+    "start-hot": "cross-env HOT=1 NODE_ENV=development electron -r babel-register -r babel-polyfill ./app/main.development",
     "package": "cross-env NODE_ENV=production node -r babel-register package.js",
     "package-all": "npm run package -- --all",
-    "postinstall": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json",
-    "dev": "concurrently --kill-others \"npm run hot-server\" \"npm run start-hot\""
+    "postinstall": "concurrently \"install-app-deps\" \"node node_modules/fbjs-scripts/node/check-dev-engines.js package.json\"",
```
pushtheworldllc: what is `install-app-debs` and where is it located?

pushtheworldllc: Also i have another post install step, could i just add it to the list?

pushtheworldllc:
I'm working in these updates to build a  native node package `openbci` in the main process.

package.json
```diff
@@ -14,11 +14,51 @@
     "build-renderer": "cross-env NODE_ENV=production node -r babel-register ./node_modules/webpack/bin/webpack --config webpack.config.production.js --progress --profile --colors",
     "build": "npm run build-main && npm run build-renderer",
     "start": "cross-env NODE_ENV=production electron ./",
-    "start-hot": "cross-env HOT=1 NODE_ENV=development electron -r babel-register -r babel-polyfill ./main.development",
+    "start-hot": "cross-env HOT=1 NODE_ENV=development electron -r babel-register -r babel-polyfill ./app/main.development",
     "package": "cross-env NODE_ENV=production node -r babel-register package.js",
     "package-all": "npm run package -- --all",
```
pushtheworldllc: This does not seem to work:
screenshot 2016-08-09 16 29 15

package.json
xwartz: `install-app-deps` is `electron-builder` script used to automatically install application dependencies. See  [install-app-deps.ts](https://github.com/electron-userland/electron-builder/blob/master/src/install-app-deps.ts)

package.json
xwartz: `release` directory... It looks like you're running script: `npm run package`.

pushtheworldllc: `npm run package` should be removed or work right?

package.json
pushtheworldllc: is this where you would recommend native, os specific, npm packages go?

pushtheworldllc:
Should we change this line https://github.com/chentsulin/electron-react-boilerplate/pull/310/files#diff-49fb32b6d8276ecd1268e47c0339d55cR28 to say `static`?

dustintownsend:
@aj-ptw I merged in @xwartz changes that appear to fix the issue with the package script. 
I'll also merged in the changes from the master. I ran into an eslint issue with the CI, but was able to solve it by updating the ignore - not sure if that is correct thing to do or not.

In my opinion all modules that aren't for development should be in the "app package.json (./static/package.json) It's not currently that way. But I believe all native module will absolutely need to be in the app package.json for electron-builder to build them correctly.

aj-ptw:
@dustintownsend Could you advise on the workflow for a native module? 
So the only way i can require modules in the new `static/package.json` from within `app/main.development.js` is like thus:
```
const OpenBCIBoard = require('../static/node_modules/openbci');
```
where the `openbci` module is installed properly. 

This works for `npm run dev` however I'm still running into problems trying to add `openbci` to the externals list in `webpack.config.electron.js`:
```
externals: [
        'openbci'
    ]
```
does not work... I've tried a bunch of absolute references too but can't get the package to not be packed because i still see this:

```
ERROR in ./static/main.js from UglifyJs
SyntaxError: Unexpected token: punc ()) [./static/~/openbci/openBCIBoard.js:197,0]
```

aj-ptw:
Why do we not follow `electron-builder` and place the other package.json in `app` dir and put it in `static`? just curious on this one

dustintownsend:
@aj-ptw I think I figured it out.

First the proper workflow should be using a normal `require(modulename)` and not having the figure out the path. So, for that to work the dependencies in the `static/node_modules` folder need to be manually loaded during development. See commit 3d9362238a19f9263629555f4587fe48faea23ca. That should work for both `npm run dev` and `npm run build`.

I did a quick test with `openbci` and it seem to work. I used their simulator example and printed out something to the console.log on the board ready event.

The reason the second `package.json` needs to be in the `static` folder instead of the `app` folder is the files in the `app` folder need to be transpiled to normal javascript and output to the `static` folder. So in this case the `app` folder is the same as the `src` folder in the `electron-builder` example. `electron-builder` will use the `static` folder for packaging.

aj-ptw:
@dustintownsend How do the source files get packaged into the final build? 
i.e. the contents of `./app`

```
new HtmlWebpackPlugin({
            filename: 'app.html',
            template: 'app/app.html',
            inject: false
        })
```
the above code does move the app.html file to the `./static` folder however this file references files in the `./app` folder when it came from, therefore in `production`, those html files cannot be found.

aj-ptw:
@dustintownsend  Ok finally some success!!

In the `web pack.config.electron.js` file i had to add:
```
module: {
  noParse: [/serialport/]
},
```
This will avoid parsing `serialport`, a module with native dependencies.

In `webpack.config.production.js` i had to do:
```
new CopyWebpackPlugin([
            {
                from: 'app/app',
                to: 'app'
            }
        ], {
            ignore: [
                // Doesn't copy any files with a js or scss extension
                '*.js', '*.scss'
            ]
        }),
        new CopyWebpackPlugin([
            {
                from: 'app/components',
                to: 'components'
            }
        ], {
            ignore: [
                // Doesn't copy any files with a js or scss extension
                '*.js', '*.scss'
            ]
        })
```
to get all my html files into the static folder. 

I did move most of the packages over to `./static/package.json`

dustintownsend:
@aj-ptw I am still learning about the magic behind webpack and don't fully understand it all. 

Awesome! I have some questions for you...

1. In the `webpack.config.electron.js` file did you include serialport in the externals prop as well or just in module?
2. In the `webpack.config.production.js` file did you delete the `HtmlWebpackPlugin`?

aj-ptw:
The mysteries of web pack are slow to be revealed!

1. In `web pack.config.electron.js` i *only* include `openbci`. Without `openbci` I get this error with a `npm run build-main`:
screenshot 2016-08-18 14 32 08
with `openbci` in the externals list, no errors
2. No I had to keep it as:
```
new HtmlWebpackPlugin({
            filename: 'app.html',
            template: 'app/app.html',
            inject: false
}),
```

aj-ptw:
I think we might need to so some work with `npm run builder:win` because there is an issue with `remoteReleases` being set by default to the github repo url

Reference: `remoteReleases` 
https://github.com/electron-userland/electron-builder/wiki/Options#buildwin

The error i'm seeing:
```
Building Squirrel.Windows installer

remoteReleases is set to https://github.com/pushtheworld/Thinker
Error: mono exited with code 255
Output:
Trying to download RELEASES index from https://github.com/pushtheworld/Thinker/RELEASES
Trying to download RELEASES index from https://github.com/pushtheworld/Thinker/RELEASES
Trying to download RELEASES index from https://github.com/pushtheworld/Thinker/RELEASES
Trying to download RELEASES index from https://github.com/pushtheworld/Thinker/RELEASES

Error output:
Attempting to sync URL as remote RELEASES folder
Failed to sync URL as GitHub repo: repos/pushtheworld/Thinker/releases was not found.
```

app/main.development.js
```diff
@@ -12,6 +12,9 @@ if (process.env.NODE_ENV === 'production') {
 
 if (process.env.NODE_ENV === 'development') {
   require('electron-debug')(); // eslint-disable-line global-require
+  const path = require('path'); // eslint-disable-line
+  const p = path.join(__dirname, '..', 'static', 'node_modules'); // eslint-disable-line
+  require('module').globalPaths.push(p); // eslint-disable-line
```
aj-ptw: This was supposed to fix the "module not found" issue in `dev` mode (`npm run dev`) when a module is in the `static/node_modules` folder? Still having a problem with that on my end

aj-ptw: Found [another project following the same work flow this way](https://github.com/sqlectron/sqlectron-gui/blob/master/src/browser/main.js#L11) from [this conversation](https://github.com/electron-userland/electron-builder/issues/326)

aj-ptw:
Here is an example of generating a `package.json` on the fly! Pretty cool

dustintownsend:
@aj-ptw for the building windows on osx, do you have wine setup per electron-builder instructions? I see that same message when I build for windows, but it does work on windows so not sure if that is an issue or not...

For remote releases I think you can delete that or set it to false if your repo isn't setup for that. I don't use that since my project is closed source and actually on bitbucket since that is what we use internally, so I haven't look into that option.

For the node_modules folder. If Thinker_New is the root folder where your first/dev package.json file is located then it should have a node_modules folder there as well as in static.
Based on that webpack link it will get any folder that is named node_modules, so that should include the one in the root and static folders.

I don't see your example for generating the package.json on the fly.

amilajack:
@chentsulin @jhen0409 Thoughts on this?

jefffriesen:
For another reference implementation and discussion if using electron-builder, check out this: https://github.com/szwacz/electron-boilerplate/issues/172

amilajack:
I have done the same for my own project:

http://github.com/amilajack/popcorn-desktop-experimental

`npm run package` will run `electron-builder` instead of `electron-packager`

amilajack:
@dustintownsend can you please rebase this PR?

PR #321, opened 2016-08-02
Install eslint v3 and plugins

chentsulin:
#320, #319 , #318 , #317, #287 

Need fix: `import/no-extraneous-dependencies`


amilajack:
What do you think about turning off this specific config?

chentsulin:
Ok. Already disable it, so should pass the linter now.

PR #323, opened 2016-08-03
Upgrade mocha to v3

jhen0409:
Also remove co-mocha. Fixes #316.


jhen0409:
The Travis build have the same error with #316, I think the problem related to `node_modules` cache.

chentsulin:
Close #316

PR #343, opened 2016-08-17
Use `electron` package name instead of `electron-prebuilt`

jhen0409:
See https://github.com/electron-userland/electron-prebuilt/issues/160.

But we should waiting https://github.com/sindresorhus/electron-is-dev/pull/3, it's dependency of `electron-debug`.


sindresorhus:
electron-is-dev 0.1.2 is published :)

PR #353, opened 2016-08-21
Remove IIFEs from app.html.

knpwrs:
Obviously up to you if you want to merge this. A neat side effect of block-scoping with `const` and `let` is that you don't need [IIFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression)s.


amilajack:
LGTM

PR #365, opened 2016-08-29
Code styles changes to configStore

amilajack:
This seems to be a bit cleaner


chentsulin:
No much different between those two style, but I am not sure that dead code elimination works with ternary operator, so I prefer just don't merge this.

PR #376, opened 2016-09-05
Adding optional flowtype annotation checks with Babel

jefffriesen:
I've really enjoyed adding Flow type annotations to build apps (https://flowtype.org). It saves me from so many little errors and it helps document functions by specifying their argument types and return types.

Here is an example:
```
function add(num1: number, num2: number): number {
  return num1 + num2;
}
const x: number = add(3, '0');
console.log('x: ', x);   
```
The above example returns 30 instead of 0. Flow will catch things like this. Flow is becoming very powerful, even allowing you to specify possible values for a argument or return value.  This PR allows people to gradually add type annotations if they want it. If they don't use or know about type annotations, they don't have to change anything. Types annotations are only checked during dev, not production.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,13 @@
 {
   "presets": ["es2015", "stage-0", "react"],
-  "plugins": ["add-module-exports"],
+  "plugins": [
+    "add-module-exports",
+    ["typecheck", {
+      "disable": {
+        "production": true
+      }
+    }]
+  ],
   "env": {
     "production": {
       "presets": ["react-optimize"],
--- a/.babelrc
+++ b/.babelrc
@@ -1,22 +1,14 @@
 {
   "presets": ["es2015", "stage-0", "react"],
-  "plugins": [
-    "add-module-exports",
-    ["typecheck", {
-      "disable": {
-        "production": true
-      }
-    }]
-  ],
+  "plugins": ["add-module-exports"],
   "env": {
     "production": {
       "presets": ["react-optimize"],
-      "plugins": [
-        "babel-plugin-dev-expression"
-      ]
+      "plugins": ["babel-plugin-dev-expression"]
     },
     "development": {
-      "presets": ["react-hmre"]
+      "presets": ["react-hmre"],
+      "plugins": ["tcomb"]
     },
     "test": {
       "plugins": [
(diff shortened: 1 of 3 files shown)


amilajack:
Use [tcomb](https://github.com/gcanti/babel-plugin-tcomb) instead of `typecheck`. It isn't [actively maintained and has a lot of issues](https://github.com/codemix/babel-plugin-typecheck/issues/172).

jefffriesen:
@amilajack Ok, good call. I'll look into it and update this PR.

amilajack:
Thanks!

jefffriesen:
@amilajack I looked at tcomb's implementation and it looks pretty good and well maintained. I switched this PR to use that plugin in dev mode. Take a look at let me know if this setup looks right to you. Also - any reason `tcomb` is suggested as a dependency and not a dev-dependency? https://github.com/gcanti/babel-plugin-tcomb

Seems like if we are only type-checking in dev mode, it could be a dev-dependency.

amilajack:
Both should be devDeps. We never run tcomb in production so it's not necessary.

amilajack:
Made a PR that fixed that:

https://github.com/gcanti/babel-plugin-tcomb#setup

amilajack:
I wrote an [eslint plugin](https://github.com/amilajack/eslint-plugin-flowtype-errors) that formats flowtype errors as eslint errors. Willing to give it a go and add it to this PR?

jefffriesen:
That eslint plugin looks interesting. I'll give it a shot in the next few days.

amilajack:
Here's a quick demo:

ESLint Flow Demo

@chentsulin @jhen0409 Thoughts on flow integration?

amilajack:
I think it would be good to add support for `eslint-pluginf-flowtype`

jefffriesen:
I'm all for it. I just have a deadline right now and can't implement it. Feel free if you want to.

amilajack:
@jefffriesen can you enable the option that lets me make edits to this PR? That would be really helpful. Thanks!

amilajack:
Closing this PR and opening a new PR because I can't read/write.

jefffriesen:
that works

Issue #375, opened 2016-09-05
async await working in main.development.js but not inside `app`


jefffriesen:
I see you have async/await functions working here:
https://github.com/jefffriesen/electron-react-boilerplate/blob/async-test/main.development.js#L35

When I try to use async/await inside the main app I get this error: `Uncaught ReferenceError: regeneratorRuntime is not defined`
<img width="461" alt="screen shot 2016-09-05 at 11 52 35 am" src="https://cloud.githubusercontent.com/assets/445733/18254844/4943ad38-735f-11e6-951f-a0f527a880d8.png">

Here is my async/await example inside index.js:
```
function sleep(ms = 0) {
  return new Promise(r => setTimeout(r, ms));
}
async function asyncTest() {
  console.log('a');
  await sleep(3000);
  console.log('b');
}
asyncTest();
```
Which works when I run it on [Babel's site](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&experimental=true&loose=false&spec=false&code=function%20sleep(ms%20%3D%200)%20%7B%0A%20%20return%20new%20Promise(r%20%3D%3E%20setTimeout(r%2C%20ms))%3B%0A%7D%0Aasync%20function%20asyncTest()%20%7B%0A%20%20console.log('a')%3B%0A%20%20await%20sleep(3000)%3B%0A%20%20console.log('b')%3B%0A%7D%0AasyncTest()%3B&playground=false)

Here is the full commit from master forked today: https://github.com/jefffriesen/electron-react-boilerplate/commit/a8fd73d80e35161183796a1e18eb73fbeb80afbb

Any suggestions appreciated. Thanks in advance.

vinnymac:
@jefffriesen if you want to use async/await inside the main app what I do is add 'babel-polyfill' as an entry point. You'll want to modify both the production and development webpack configurations.

In Development
```javascript
  entry: [
    `webpack-hot-middleware/client?path=http://localhost:${PORT}/__webpack_hmr`,
    'babel-polyfill',
    './app/index'
  ],
```

In Production
```javascript


  entry: [
    'babel-polyfill',
    './app/index'
  ],
```

jefffriesen:
Ok, good call. Thank you. Do you think other people would want this? I can do a PR for it.

jefffriesen:
@vinnymac Works great. I submitted a PR if other folks are interested in having this.

[...]

PR #378, opened 2016-09-07
Enabling use of async/await in renderer

jefffriesen:
This adds the babel-polyfill to the entry of dev and production webpack configs to allow people to use async/await in their app. 

For an example async/await function, add this anywhere in the renderer code:
```
function sleep(ms = 0) {
  return new Promise(r => setTimeout(r, ms));
}
async function asyncTest() {
  console.log('a');
  await sleep(3000);
  console.log('b');
}
asyncTest();
```

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -10,6 +10,17 @@ import './app.global.css';
 const store = configureStore();
 const history = syncHistoryWithStore(hashHistory, store);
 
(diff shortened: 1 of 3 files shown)


jefffriesen:
Related to ticket #375

jhen0409:
Thanks! :)

PR #387, opened 2016-09-14
Use handlebars for main app page.

knpwrs:
Just a thought. If this were to be merged there may be a discussion around how much logic you want in the template vs in the variables being passed to the template.


amilajack:
why?

knpwrs:
It's just a different way to do things. It doesn't really matter. I only found out about the APIs for intercepting the file protocol earlier today and thought it was cool (after seeing applications for pug and ejs templates).

amilajack:
I don't really think this dependency is worth its weight. Its better to use idiomatic JS IMO.

knpwrs:
Fair enough. Probably fits better in an application that already uses Handlebars.

amilajack:
Yup, exactly what I was thinking!

amilajack:
Should this be closed then?

knpwrs:
May as well. I typically let the repo owner weigh in.

PR #392, opened 2016-09-17
Changed 'node' webpack config to 'test'

amilajack:
I think this makes more sense. Thoughts?


jhen0409:
LGTM, but I prefer to use `test`. @chentsulin what do you think?

amilajack:
Good idea. I'll change to `test`

chentsulin:
prefer `test` too.

amilajack:
Just changed

PR #402, opened 2016-09-20
add cleanup components

tstringer:
*Extremely* huge fan, and frequent user of this boilerplate.  So first off, thank you for creating and maintaining this.

As a frequent user of the boilerplate, I had to remove a lot of the (necessary and appropriate) sample code for the boilerplate (mostly surrounding the counter components).  I think this is necessary so that you can quickly and easily see a working application.

But I wanted a way to be able to quickly and automatically clean up the boilerplate so that I can start custom dev as soon as possible.  So I created [`boiler-room-custodian`](https://www.npmjs.com/package/boiler-room-custodian) ([GitHub repo](https://github.com/tstringer/boiler-room-custodian)), which is a little utility that allows boilerplate maintainers to specify a configuration for cleanup, and then run `mop` to do the cleanup.

In the case of this specific boilerplate, I wanted to use it not only as a testing ground for this utility but also as a constant user of this boilerplate I benefited from creating the config.

Let me know your thoughts and what value you think this could bring to the boilerplate.  If this PR is merged, users would only need to run `npm run cleanup` to remove the sample code and then start custom development.

Thank you for your time and consideration!  And thanks again for your hard work on this boilerplate!


amilajack:
I think changing the filename `cleanup.js` to `boiler-room-custodian.js` would be better.

tstringer:
Ok I can see why you think that, and I'm not against changing it away from `cleanup.js`.  I think `boiler-room-custodian.js` would be a bit of a long string for any filename, much less a config filename.  How about `brcustodian.js`?  Or `brc.js`?  Let me know your thoughts.  I'll make the necessary changes on the module as it is hard-coded to look for `cleanup.js`.

Thanks for the input! 👍  😄

amilajack:
How about`br-custodian.js`? `brcustodian.js` is less readable IMO

tstringer:
Ok yeah I can see how `brcustodian.js` is bad.  I don't like the "remove" in `remove-boilerplate.js`, as there are modifications and even file adds that happen.  Maybe `cleanup-boilerplate.js`?  Or `sanitize-boilerplate.js`?  Not sure I love "sanitize" though.  If neither of these sound good to you, I'd be willing to go with `boiler-room-custodian.js` as a fall back just to have something finalized. :smile:

Let me know your thoughts!

amilajack:
So here's my judgement on the naming.

When I first used this project, I was really confused about how the project worked. "Why is this this way? What does this do?" Those are the kinds of questions that I feel a lot of other people experience with this project and boilerplates in general.

A solution to this is adding comments and explanations to each significant file. Users should feel they know their way around the project even without having much experience with the technologies that this project uses.

Another solution is, as trivial as it might sound, is the naming of files. The user should know what that file does just by reading the filename.

How about a more general filename, like `bootstrap.js`?

tstringer:
Totally agreed! My only fear of "bootstrap" is that they'll confuse it with the popular bootstrap library. That's why I originally thought "cleanup" was pretty descriptive. How about just "sanitize.js"? Or "square-one.js". Or "minimize.js"? Thanks again :-)

amilajack:
`setup.js`?

tstringer:
That works for me! I'll make the code changes tomorrow, and they'll reflect in this pull request.

Thanks for the recommendation!! :smile:

amilajack:
Great! And sorry for being so picky!

tstringer:
No need to apologize at all, naming is so hard and frustrating. 😄 

Changes made, let me know your thoughts!

amilajack:
LGTM! Merging 🎉

tstringer:
Awesome, thanks!!! :+1: :smile:

amilajack:
Also here's something that would be a great addition to boilerplate custodian: removing boilerplate comments. I'm planning on adding comments to this another boilerplates, including `react-boilerplate` and others. Comments are needed to explain the decisions of the boilerplate and how things work. Ideally, the comments would be removed automatically with a single script, like boilerplate-custodian. Thoughts?

tstringer:
That's a really good idea!  How would you like to see that implemented?  Removing *all* comments in a specified file?  Or just specified comments?

amilajack:
How about something like this:

```js
/**
 * Some Boilerplate Here!
 * An example of a comment that should be removed
 * To make it easier to distinguish, we could add an annotation like so:
 * @boilerplate
 */

/**
 * Example of a comment that won't get deleted because it doesnt have a `@boilerplate` annotation
 */
```

tstringer:
I love it ❤️  I'll integrate that functionality!  Super great suggestion. 👍

PR #408, opened 2016-09-21
Added hot reloading for functional components and refactored dumb components to functional style

sdeleon28:
Added hot reloading for functional components and refactored dumb components to functional style

### DIFF
--- a/app/containers/HomePage.js
+++ b/app/containers/HomePage.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
 import Home from '../components/Home';
 
 export default () => (
--- a/app/containers/HomePage.js
+++ b/app/containers/HomePage.js
@@ -1,4 +1,4 @@
 import React from 'react';
 import Home from '../components/Home';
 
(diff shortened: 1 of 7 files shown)


app/containers/HomePage.js
```diff
@@ -1,10 +1,6 @@
...
 import Home from '../components/Home';
 
-export default class HomePage extends Component {
-  render() {
-    return (
-      <Home />
-    );
-  }
-}
+export default () => (
```
amilajack: Wouldn't this be better?

```export default () => <Home />
```

amilajack: Wouldn't this be better?

```js
export default () => <Home />
```

sdeleon28: Sure, if we add a semicolon ;-) I'll do that in a sec

sdeleon28: Looks like one of the tests times out without the parentheses in Node 5. I'll try with the parentheses but inline.

chentsulin:
Superseded by #421

PR #418, opened 2016-09-24
Making HMR to work on Windows

andrerpena:
HMR not working on Windows and it took me like 40 minutes to figure it out. I verified this was the cause before sending this PR.


chentsulin:
@amilajack babel-loader will retrieve config from `.babelrc` based on `NODE_ENV` env variable, but windows must use different way to set env variables. See: https://github.com/kentcdodds/cross-env#why

amilajack:
I see. Thanks for clarifying

PR #421, opened 2016-09-25
Replace react-transform with react-hot-loader v3

chentsulin:


amilajack:
@chentsulin @jhen0409 Just wondering. Why wasnt this merged yet? Is there some upstream bug that we're waiting on to be fixed?

amilajack:
Migration for this was finished

Issue #385, opened 2016-09-12
window is undefined


obecny:
I have an electron app build on top of  "electron-react-boilerplate". When I try to use build and then run I'm getting an error "window is undefined".
After some investigation it turn out that the problem is related with package "form-data"

"The easy fix" is to simply remove browser param from package.json for "form-data" in node-modules.

Now the main question is that somehow related to form-data itself  or a webpack or the configuration for webpack in this project.
I have already post the same question in "form-data" repo
https://github.com/form-data/form-data/issues/252

Any idea how this could be solved in a proper way would be really useful,
thx :)

amilajack:
You should fork the repo, remove the browser field from your fork, and add it as a github dep:

```
dependencies: {
  form-data: 'username/form-data'
}
```

If the package requires compilation, try adding the build directories to version control

Another workaround is creating a temporary npm package.

Let me know how this works out.

amilajack:
Here's an example:

https://github.com/amilajack/popcorn-time-desktop/blob/master/package.json#L174

jaxgeller:
There's probably a webpack option like `externals` or something else around this, but in the mean time I've created a fork here `https://github.com/jaxgeller/form-data` you can use it like this `"form-data": "github:jaxgeller/form-data"` in your package.json.

worldjoe:
I have this problem too. The underlying problem is that Webpack doesn't fully understand the electron nor electron-main when trying to figure out if the target is web or not.

One fix is to add this to webpack.config.electron.js:

`  resolve: {
    packageAlias: "main"
  },`

Another fix appears to be to bump the webpack version. Pull request coming.

worldjoe:
I'm testing this fix out. I cleaned my node_modules folder and the fix doesn't seem to work anymore. Not sure if it's dependent on a specific version of webpack. But I think the root cause is still this. webpack (1.x) defaults to browser when packageAlias is not set.

obecny:
@worldjoe indeed , doesn't work for me either :(

obecny:
@worldjoe so far the fork and removing the browser from package.json only works, hopefully there is going to be a better one

amilajack:
I believe this is a webpack issue or an issue with the package itself. I had to do this for my own projects that use this boilerplate.

worldjoe:
The problem is in webpack.

https://github.com/webpack/webpack/blob/webpack-1/lib/WebpackOptionsDefaulter.js#L75

[...]

PR #431, opened 2016-09-29
fix window is undefined #385

ArnaudValensi:


amilajack:
I can't figure out why this fixes it. Can you please explain this. Thanks!

ArnaudValensi:
The modification was originally suggested by @worldjoe [here](https://github.com/chentsulin/electron-react-boilerplate/issues/385#issuecomment-249276167)

Here is my understanding:

When we create a npm module, in the `package.json` we can add different entry points to use a module in different contexts, like as node module, browser, etc.

Here is the `package.json` of `form-data`:
```
...
  "main": "./lib/form_data",
  "browser": "./lib/browser",
...
```

`resolve.packageMains` seems to tell webpack what entry point pack.

amilajack:
I see. Have you tried testing this on your local machine? Also I think it would be better if you could merge this into the new `dev-master` branch so that I can test this out. Thanks!

ArnaudValensi:
Yes, I tried it. Actually I had the issue and I was looking for a solution to put my software in production.

ArnaudValensi:
@amilajack I don't see any `dev-master` branch.

amilajack:
Sorry about that. Forgot to push from my local env. You should see it now.

ArnaudValensi:
I've made another PR into dev-master here: https://github.com/chentsulin/electron-react-boilerplate/pull/436

Issue #266, opened 2016-07-06
responseCallback is not supported


AlexFrazer:
Getting this error in my console when starting up the application from scratch without having changed anything

jhen0409:
It's from [redux-devtools-extension](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/src/browser/extension/options/syncOptions.js#L52), also related to https://github.com/electron/electron/issues/6307. We didn't use the options page (it's not supported), so it's no problem. Maybe we should avoid this error.

jhen0409:
It should solved by [redux-devtools-extension@2.3.0](https://github.com/zalmoxisus/redux-devtools-extension/releases/tag/v2.3.0).

burgerzzz:
I'm still getting this error after pulling down latest release

jhen0409:
@burgerzzz could you check the extension version on the console? Just run:

```js
require('electron').remote.BrowserWindow.getDevToolsExtensions()['Redux DevTools'].version
```

It should no any errors if extension version >= 2.3.0, because I have write a test to ensure no output logs in [redux-devtools-extension repo](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/test/electron/devpanel.spec.js#L86).

burgerzzz:
Looks like you were correct, in that I have version "2.2.1.1" of redux-devtools. Is there a reason electron-devtools-installer installed that version, or is it installed globally from that package? Guess, I'm not sure how to update redux-devtools to >=2.3.0.

jhen0409:
@burgerzzz try use `UPGRADE_EXTENSION=1 npm run dev` to upgrade extensions (I'll add it to readme), if you're tried and it's still getting old version, it maybe related to https://github.com/GPMDP/electron-devtools-installer/pull/14, please wait new version of devtools-installer, or just run following code on the console and restart app:

```js
require('electron').remote.BrowserWindow.removeDevToolsExtension('Redux DevTools')
```

PR #440, opened 2016-10-06
Use electron-builder, continued

zeevl:
This is a rebase of #309 as requested.    Not sure if @dustintownsend is still active on this, but it has been a while and it's become something I need so I'm joining in the quest!


amilajack:
Thanks for this! 🎉

amilajack:
The `regeneratorRuntime` error is caused by not including [babel-regenerator](http://babeljs.io/docs/plugins/transform-regenerator/), which is a production and runtime dependency of babel for certain babel polyfills.

zeevl:
Well, that's annoying.. travis fails on node 4 (chromedriver timeout) but succeeds on node 6, while appveyor fails on node 06 and succeeds on node 4.  Any ideas?

dustintownsend:
@zeevl awesome! Thanks for picking up my slack. I was on vacation, so I haven't had much time lately. I'll try to help out soon after I get caught up on everything at work.

dustintownsend:
@zeevl one thing I noticed is the font awesome isn't getting loaded. You need to update the path in app.html. I don't think it needs a different path for hot loading. 

You could just remove `const font` and replace `link.href` line with this:
`link.href = './node_modules/font-awesome/css/font-awesome.min.css';`

amilajack:
@zeevl I believe [this](https://github.com/amilajack/eslint-plugin-flowtype-errors/blob/master/appveyor.yml#L18) is how you force appveyor to use x64 arch

zeevl:
Ah cool, thanks @amilajack and @dustintownsend

amilajack:
Yay! All tests passing. Is this ready to be merged?

zeevl:
It's working well for me.   Do you want a version bump?

amilajack:
>  Do you want a version bump?

In the `package.json`? I think this should be done with the approval of @chentsulin. Also I don't have npm publishing permission.

zeevl:
Right on thanks.  Let me know if there's anything else you need from me!

.gitignore
```diff
@@ -25,12 +25,16 @@ build/Release
...
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 node_modules
+app/node_modules
 
 # OSX
 .DS_Store
 
 # App packaged
 dist
-release
```
amilajack: If `release` is no longer used, where are the built folders going? Ideally, they we could build to the same folder. A lot of users of this project already expect builds to be there so changing the directory would not be a  good move.

[Here's how you fix that](https://github.com/amilajack/popcorn-time-desktop/blob/985b85c2615f6245da7480afd3102980194a2ce3/package.json#L63)

amilajack:
@zeevl Totally forgot to add this to my review but I think you should how to add icons to the packaged app. I just remembered [this issue](https://github.com/chentsulin/electron-react-boilerplate/issues/407)

Also I think we should change the 'build' director to be 'resources'. 'build' sounds like a place where compiled files would go.

```
```

zeevl:
Yeah I wasn't a fan of the `build` directory either, however I'm guessing that was chosen as it's the default for icons in `electron-builder`.   They even say [dont customize paths to the background and icon, just follow conventions](https://github.com/electron-userland/electron-builder/wiki/Options#options).   I'll look into changing it nonetheless.

One question for you @amilajack, I noticed when running the windows installer, an initial splash screen opens, then the application opens a few seconds later *while the splash screen is still up*, then when the install completes the application closes then reopens again.  This doesn't seem to happen in [your repo](https://github.com/amilajack/popcorn-time-desktop) though.   Does this sound familiar and is there something you did to fix it?

amilajack:
Yup! That's the splash screen:

screen shot 2016-10-08 at 1 50 03 pm

https://github.com/electron/windows-installer/blob/master/resources/install-spinner.gif

Not sure how I fixed it. 

> They even say dont customize paths to the background and icon, just follow conventions. I'll look into changing it nonetheless.

Great! We'll see if we have any issues with adding custom icons. If we do, we can revert to the default.

zeevl:
Icons were already created and committed by @dustintownsend in #310

amilajack:
Build is failing because of linter errors in `app/node_modules`. Not sure why this is happening because I'm pretty sure that all files in `.gitingore` are ignore by eslint.

zeevl:
Figured out why the application was running multiple times during install -- https://github.com/electron/windows-installer#handling-squirrel-events

https://github.com/chentsulin/electron-react-boilerplate/pull/440/commits/8bde40c0afb3620ce860229566834f389505ccf3 fixes it

zeevl:
linting: see https://github.com/eslint/eslint/issues/5848

amilajack:
LGTM! @chentsulin @dustintownsend @jhen0409 would be great if you could review as well in case i missed anything

amilajack:
@zeevl btw did you try manually packaging on all os's? Even tho the tests show that the packaging is successful, it is very likely that they do not package correctly.

amilajack:
Icon seems a bit small

screen shot 2016-10-08 at 6 43 25 pm

develar:
FYI: In the next release of electron-builder Squirrel.Windows will be not default target for windows. NSIS will be used instead. https://github.com/electron-userland/electron-builder/issues/529 is implemented and will be tested/closed next week. electron-boilerplate already uses NSIS by default (https://github.com/szwacz/electron-boilerplate/issues/172#issuecomment-232618357).

package.json
```diff
@@ -106,20 +154,6 @@
     "webpack-merge": "^0.14.1",
     "webpack-validator": "^2.2.9"
   },
-  "dependencies": {
```
jhen0409: I think we should keep dependencies, we can clearly know which are main dependencies, and which are development tools. It doesn't affect the electron-builder.

Also, `css-modules-require-hook`, `postcss` seems unnecessary.

app/app.html
```diff
@@ -3,23 +3,27 @@
...
-        document.write(link.outerHTML);
-      }
+      (function() {
+        if (!process.env.HOT) {
+          const link = document.createElement('link');
+          link.rel = 'stylesheet';
+          link.href = './style.css';
+          document.write(link.outerHTML);
+        }
+      }());
```
jhen0409: See #353.

package.json
zeevl: @jhen0409 how would you explain difference between what should be in `devDependencies`, and what goes in `dependencies` in `./package.json'?    FWIU anything that gets packaged with an app goes in `./app/package.json`; everything else belongs in the root `./package.json`.     And as such, it makes sense to me that `./package.json` only has a `devDependencies`..

amilajack:
> FYI: In the next release of electron-builder Squirrel.Windows will be not default target for windows.

@develar Thanks for mentioning this! @zeevl would be great if you could add this as the default so uses can don't have to worry about migration themselves.

package.json
jhen0409: @zeevl I'm used and agreed [this ESLint rule](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L69) (Although it's disabled here), it's the reason for why I distinguish main dependencies and development tools. (Of course, exclude dev packages in `development` node env)

> FWIU anything that gets packaged with an app goes in ./app/package.json; everything else belongs in the root ./package.json

I prefer just put webpack externals to dependencies of `./app/package.json`, we can focus on __native__ modules, sometimes I can only add deeper dependencies to reduce bundle size, so I think keep dependencies of `./package.json` not bad, 
it is not related to production build.

develar:
>  I think we need to set AppImage as the default

It is already default in the current electron-builder version.

package.json
amilajack: LGTM. Waiting for others to review.

zeevl:
@amilajack 
> btw did you try manually packaging on all os's?

No.. I only have access to macOS at the moment.   Is it possible someone else could test building on windows & linux?

> Icon seems a bit small

Is that a linux platform?   again, no easy access to that right now so I'm not even sure where electron-builder is grabbing that..

amilajack:
> Is that a linux platform? again, no easy access to that right now so I'm not even sure where electron-builder is grabbing that..

I was on Mac OS Sierra

zeevl:
> I was on Mac OS Sierra 

Ah ok, the About dialog.   Never looked at that before. :/

Regenerated the icns file, seems to have fixed it.  No idea why.

amilajack:
@zeevl I'm manually check this on windows and linux later today and get back to you.

dustintownsend:
screenshot 2016-10-17 11 07 55

Anyone else getting this error in the dev tools console? `responseCallback is not supported  chrome-api.js:124`

amilajack:
@dustintownsend I havent gotten this.

jhen0409:
> Anyone else getting this error in the dev tools console? responseCallback is not supported chrome-api.js:124

@dustintownsend, this is because you're using old version of `redux-devtools-extension` (See #266), you can use `UPGRADE_EXTENSIONS=1 npm run dev` to upgrade extensions. (See [README](https://github.com/chentsulin/electron-react-boilerplate#devtools-extension))

dustintownsend:
@amilajack @jhen0409 okay good to know it's just something with my machine. I tried that, but didn't work. Since it's unrelated to the boilerplate I'll open an issue with the devtools if I'm unable to figure it out. Thanks.

amilajack:
@dustintownsend What machine did you test this on? Its good if we could test all OS's before merging. I've tested on mac and worked fine.

dustintownsend:
@amilajack 

> Is this after or before packaging?
It was happening before, in development mode. npm run dev
I'm not sure if it's possible to run the dev tools on the packaged build.

@amilajack I'm running Mac as well, OS X El Capitan. What needs to be tested?

amilajack:
@dustintownsend we have yet to test on linux and windows

develar:
> I'm not sure if it's possible to run the dev tools on the packaged build.

It is possible. But as far I know, only JetBrains IDE supports ASAR archives (WebStorm, IntelliJ IDEA).

dustintownsend:
I did some quick testing on Windows 10 (VMWare Fusion)

The only issue I have run into is with `npm run dev` command. I get a bunch of `Expected linebreaks to be LF but found CRLF.` messages. The shell loads with the dev tools, but just a blank screen. The dev tools console says `GET http://localhost:3000/dist/bundle.js 404 (Not Found)`.
This could be a setting on my machine with line ending. I did update the global git settings to not use CRLF and re-pulled the repo down and got the same thing. I can look into that more and make sure I didn't miss something.

I was able to do `npm run build` & `npm start` and the app loaded up. Clicking around everything appears to be working. I also ran 'npm test' & `npm run test-e2e` and all the test passed. Finally I did `npm run package` and it created the package for windows. I was expecting that to fail there was a separate command for package-win. The installer worked good and the program functioned as expected.

I did a build on my Mac as well and tested the unpacked and installer and all worked.

amilajack:
@dustintownsend Can you try testing `npm run dev` on the master branch?

amilajack:
@zeevl Would be great if you could fix the conflicts

amilajack:
When adding new deps, you need to run `yarn` so that dependencies are added to it.

What I do is this:

```
rm -rf node_modules
yarn
```

There's probably an more proper way of doing this.

amilajack:
Node why this is failing on appveyor node 4.

amilajack:
Seems like when fixing the merge conflicts in the `package.json`, I accidentally removed some of the necessary dependencies. Fixed that

amilajack:
@dustintownsend would be great if you could test this on windows again.

amilajack:
@jhen0409 I'm getting the error that @dustintownsend was getting:
screen shot 2016-10-29 at 9 28 22 pm

Ran `UPGRADE_EXTENSION=1 npm run dev` and that doesn't seem to fix it

jhen0409:
> `UPGRADE_EXTENSION=1`

@amilajack missing `S`? (`UPGRADE_EXTENSIONS`)

Otherwise use `require('electron').remote.BrowserWindow.getDevToolsExtensions()` to check extension version.

amilajack:
@jhen0409 Fixed. Thanks!

amilajack:
Tested on:
✅ Mac
✅ Linux
✅ Windows

⛴⛴⛴

aj-ptw:
I'm so excited this got finished and merged!! Great work all!

dustintownsend:
Awesome @amilajack! Sorry I left you hanging on testing it - I didn't see the notification.

I'm starting from scratch to test everything out as I'm migrating my existing app code into this. The only issue I've run into so far is Yarn doesn't work for me, but I see there is already issues created for that. I'll be sure to open up new issue if I run into anything.

develar:
> Yarn doesn't work for me

Please see https://github.com/electron-userland/electron-builder/issues/861#issuecomment-257221259

PR #471, opened 2016-10-21
Get electron version for build with a require rather than an exec npm

alexstrat:
For random and, most of the time, legit reasons (peer dependencies things, blabla), the `npm list electron --dev` command fails (very silently), ending up packaging the app with a totally different version than the installed prebuilt.
It's a error-prone behavior in my opinion (I got tricked).

There is 2 ways to fix it:
- make the error more noisy and make the packaging fails
- use a more robust method to get the prebuilt electron version

I chose the 2nd one here.

### DIFF
--- a/package.js
+++ b/package.js
@@ -62,15 +62,9 @@ if (version) {
   startPack();
 } else {
   // use the same version as the currently-installed electron-prebuilt
-  exec('npm list electron --dev', (err, stdout) => {
-    if (err) {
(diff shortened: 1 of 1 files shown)


jhen0409:
We are ready to merge #440, it will remove `package.js`, so this PR will not merged, sorry but this is great! :D

alexstrat:
Ya I guessed so. Definitely waiting for #440 as well :)

Issue #488, opened 2016-10-30
`npm run dev` race condition


enagy27:
Running this on OSX Sierra, npm version 3.10.9 and node version 7.0.0. After cloning and running either npm install or yarn install I'll use npm run dev, which doesn't work. Fortunately by running the two commands simultaneously I can get a dev environment working (referring to npm run hot-server and npm run start-hot), and then afterwards I can run npm run dev, but on a clean install it doesn't seem to work.

amilajack:
That isn't necessary. What is the output of running `npm run dev`?

enagy27:
Ok, so this might be a subtle difference that I wasn't expecting running on OSX my first time. What's happening is that there's a successful build and the electron window opens up, but I'm seeing ERR_CONNECTION_REFUSED because dist/bundle.js could not be loaded (not found). It seems that if I don't open the window until the build process completes- everything is reloaded and it's fine, otherwise reload just needs to be done manually.

amilajack:
Right now a manual reload is required.

amilajack:
So if I understand you correctly, the build shows an error message but when refreshing electron, everything works?

enagy27:
Nope, there are no build errors, I'm seeing the connection refused/bundle.js missing message from inside the developer tools console.

amilajack:
I see. I have to do the same thing locally sometimes. There's a race condition. Electron is opened before the build is complete.

Here's what we have right now: https://github.com/chentsulin/electron-react-boilerplate/blob/master/package.json#L21

Here's what we should have something like this:
https://github.com/amilajack/popcorn-time-desktop/blob/master/server.js#L30-L38

enagy27:
Right, yeah, makes perfect sense. Should I close this out?

amilajack:
Let's close this once I get around to fixing this.

amilajack:
Made https://github.com/chentsulin/electron-react-boilerplate/pull/489

amilajack:
Fixed

Issue #341, opened 2016-08-15
PR: Add support for webpack-dashboard


amilajack:
Willing to make a PR for migrating to [webpack-dashboard](https://github.com/FormidableLabs/webpack-dashboard). 

The primary benefit of using it is its significantly cleaner output.

Would this PR be welcome?

fearintino:
I think it will be very cool!

amilajack:
Thanks! Btw it doesnt look like travis ci is enabled:

https://travis-ci.org/JamesTheHacker/Neuron

amilajack:
@chentsulin @jhen0409 Thoughts?

chentsulin:
+1 for this, webpack bundle size analyzer should be helpful.

jhen0409:
👍  I'm all for this.

vinnymac:
I just found this after adding it to my fork. It works well.

https://github.com/vinnymac/electron-react-boilerplate/commit/501e2641bdbf2cc6931b9c420d521e53b392867a

amilajack:
@vinnymac can you show me the output of your fork? I've cloned it and started running it with no success.

vinnymac:
@amilajack all I had to do was `npm install` and `npm run hot-server`

This is what I see after hot-server starts.
![Hot Server](http://i.imgur.com/GgnQhzT.png)

node v6.2.2
npm v3.10.3

amilajack:
There are issues with webpack-dashboard on windows. `react-boilerplate` has [decided to wait until those issues are resolved](https://github.com/mxstbr/react-boilerplate/pull/886) and I think we should do the same

PR #489, opened 2016-10-30
Fixed issue #488: `npm run dev` race condition

amilajack:


amilajack:
@chentsulin @jhen0409 would be great if you could test this out locally just to be safe.

jhen0409:
I hope it can be continuously output logs of main process, so it should use `spawn`, just like [better-npm-run](https://github.com/benoror/better-npm-run/blob/master/lib/exec.js) usage.

Also, currently I still prefer to separate `hot-server` and `start-hot`, because I often change the main process, and no restart webpack server, it's even better if it can be used as an option for `hot-server`. 👍

amilajack:
@jhen0409 the main process still does output logs.

> Also, currently I still prefer to separate hot-server and start-hot, because I often change the main process, and no restart webpack server

Those commands still exist. Actually `npm run dev` still uses them.

jhen0409:
> the main process still does output logs.

No, the logs will not be output continuously if we used `child_process.exec`.

Try `setInterval(() => console.log('hey'), 100)` on main process.

> Those commands still exist. Actually npm run dev still uses them.

I mean we can provide an option for `hot-server`, currently if I just run `npm run hot-server` it will start Electron app for this PR.

amilajack:
I'm testing `spawn` locally like this:

```
```

jhen0409:
@amilajack, just try the following code, it's works well for me:

```js
spawn('npm', ['run', 'start-hot'], { stdio: 'inherit' })
  .on('close', code => {
    process.exit(code);
  })
  .on('error', _error => {
    console.error(_error);
  });
```

Because it will check `npm` command exists, not `npm run start-hot`.

Also, see https://github.com/benoror/better-npm-run/pull/24 and https://github.com/benoror/better-npm-run/pull/46, I think `stdio: 'inherit'` will be better here, unless we want to add additional characters to each output.

amilajack:
Fixed. What's the flag option that we should add? I don't understand that fully

jhen0409:
@amilajack, see https://github.com/chentsulin/electron-react-boilerplate/pull/162#issuecomment-190619469, I would like to have a way to open the webpack server only, and we have #341, so I think restart the app and keep the server running will be more convenient. I expected it on `npm run hot-server`, what do you think?

amilajack:
>  I expected it on npm run hot-server

Can you show me an example/pseudo code for this? Still not sure if I get you 100%. I know that you want to be able to restart both the webpack and the app separately. But when you say `I expected it on npm run hot-server`, I'm confused. Didn't we already do this by starting `npm run start-hot` in `server.js`, which runs when `npm run hot-server` is called?

jhen0409:
@amilajack, I mean `open webpack server *only*` on `npm run hot-server`, but this PR doesn't do, it affected [the first usage of README](https://github.com/chentsulin/electron-react-boilerplate#run).

I think we need a flag option is just like:

```json
"dev": "npm run hot-server -- --start-hot",
```

amilajack:
@jhen0409 ready to merge?

jhen0409:
OK, merged. :D

PR #491, opened 2016-10-30
Updated deps, updated gitignore

amilajack:


amilajack:
@chentsulin tests might be failing because of the cache. Can you try clearing the appveyor cache?

PR #498, opened 2016-10-31
Add flag option `start-hot` and apply env

jhen0409:
To #489

1. Add flag option `start-hot`
2. Apply env to `start-hot` mode of server.js, ensure `PORT`, `UPGRADE_EXTENSIONS` can be used for `start-hot`

/cc @amilajack

### DIFF
--- a/server.js
+++ b/server.js
@@ -36,7 +36,7 @@ const server = app.listen(PORT, 'localhost', serverError => {
   }
 
   if (argv['start-hot']) {
-    spawn('npm', ['run', 'start-hot'], { stdio: 'inherit' })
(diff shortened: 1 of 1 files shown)


amilajack:
IMO, [yargs](https://github.com/yargs/yargs) is cleaner.

jhen0409:
I think `minimist` is very suitable for small CLI tools, because it [doesn't have any dependencies](https://github.com/substack/minimist/blob/master/package.json), but all dependencies of yargs need ~1.5Mb disk space.

amilajack:
> all dependencies of yargs need ~1.5Mb disk space.

I see. Lets go with `minimist` then.

PR #499, opened 2016-10-31
[AppVeyor] Reset node_modules cache when package.json changed

jhen0409:


amilajack:
Yesss! Thanks for this! Is there a way to do this for travis?

PR #505, opened 2016-11-01
Fix port for app.html

jhen0409:
it looks #440 reverted some of the things changed. 😂


jhen0409:
It seems `html-webpack-plugin` will broken string template usage, so I temporarily back to es5 for `app.html`.

Issue #506, opened 2016-11-02
how to use setup.js


FTAndy:
I see a setup.js file in the folder, how can i use it to reset the project? What command should i use?

amilajack:
What do you mean by reset? Do you mean 'removed all the boilerplate' or 'restore it to the original version of the boilerplate'?

FTAndy:
yes

amilajack:
> yes

Which option?

FTAndy:
I mean how to use the `setup.js` file, `node setup.js` absolutely not work. Need to use any tool to execte the `setup.js` file?

amilajack:
ping @tstringer

tstringer:
@FTAndy you don't directly consume or invoke `setup.js` in any manner. You do `npm run cleanup` in order for `boiler-room-custodian` to clean up the sample code from the boilerplate.

I hope that makes sense!

FTAndy:
@tstringer thanks, then

PR #508, opened 2016-11-02
Create documentation around the cleanup script

tstringer:
There has been some confusion as to the point of `setup.js` and what it means. This PR is aimed to clarify this as well as reduce confusion.

This is in response to issue #506

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ And then install dependencies.
 $ cd your-project-name && npm install
 ```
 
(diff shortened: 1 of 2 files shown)


amilajack:
@chentsulin @jhen0409

tstringer:
@amilajack I agree with that, and we don't necessarily need to have that snippet in `README.md`. As you can see from the commit, I also included commenting in `setup.js` to explain what it is there for and how it should be used.

So I'm not against removing the snippet from the readme.

PR #517, opened 2016-11-05
Debugger fixes and improvements

amilajack:


amilajack:
Webpack script blackboxing:

https://twitter.com/apexskier/status/788134726703788032

webpack.config.development.js
```diff
@@ -15,7 +15,7 @@ const port = process.env.PORT || 3000;
 export default validate(merge(baseConfig, {
   debug: true,
 
-  devtool: 'cheap-module-eval-source-map',
+  devtool: 'inline-source-map',
```
rahilsondhi: I'm getting much faster initial build and incremental build performance by using `cheap-inline-source-map`. May I ask why you chose `inline-source-map`?

amilajack: Its a tradeoff. See https://webpack.github.io/docs/configuration.html#devtool

PR #535, opened 2016-11-11
Add parsing command line arguments

dplusic:


jhen0409:
It's helpful for any issue? Or you have some better use cases, so that you want use it by default?

dplusic:
My case is a visual merge tool like kdiff3.
The tools are usually executed with file paths in command line arguments.
This is common for other desktop applications too.

amilajack:
@dplusic would be great if you could show us how this could be used. How can compiled electron apps (dmg, exe, etc) be executed from the cli (without using npm scripts or global)?

dplusic:
cli> ElectronTextEditor.exe /path/to/txt ↲

Then, "/path/to/txt" will be passed to main.js, and the app can open the file directly.
I wonder whether I understand your question correctly.

amilajack:
Can it handle cli arguments like so?:

```
ElectronApp.app --name=Jack --age=19
``

dplusic:
Sure.

hardchor:
This really feels like it should be user-land, not part of the boilerplate.

Issue #525, opened 2016-11-08
Avoid inclusion of es6 source in final app distribution


Mike-Dax:
Issue #487 appears to be essentially this.

It doesn't seem to make sense to include the es6 source in the final app distribution.

We could set package.json to include a `build.files` array?

```
    "files": [
      "**/*",
      "!actions${/*}",
      "!api${/*}",
      "!components${/*}",
      "!containers${/*}",
      "!reducers${/*}",
      "!store${/*}",
      "!utils${/*}",
      "!main.development.js"
      "!index.js"
      "!routes.js"
    ]
```

PR #540, opened 2016-11-12
Prevent inclusion of es6 source in distribution package

Mike-Dax:
Added build.files array to prevent the es6 source from being included in the final app distribution as per issue #525.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,17 +42,12 @@
       ]
     },
     "files": [
-      "**/*",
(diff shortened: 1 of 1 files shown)


jhen0409:
LGTM, but maybe just add allow files will be better?

```json
"files": [
  "dist/",
  "app.html",
  "main.js",
  "main.js.map",
  "package.json"
]
```

PR #558, opened 2016-11-18
Adapt publicPath in webpack.config.production.js

rweindl:
Seems that the publicPath was not set correctly. This prevented referenced files (e.g. with the url(...) parameter in less files) to load correctly.


amilajack:
<3

amilajack:
Reverted because this broke how fonts are loaded. See https://github.com/chentsulin/electron-react-boilerplate/issues/559#issuecomment-261795941

PR #569, opened 2016-11-23
Webpack Image Support

amilajack:
**DO NOT MERGE. IN PROGRESS**


jhen0409:
We already have `svg` load in [renderer config](https://github.com/chentsulin/electron-react-boilerplate/blob/master/webpack.config.production.js#L50).

amilajack:
What do you think about moving these changes to the renderer instead?

amilajack:
@jhen0409 would be great if you could review

jhen0409:
> What do you think about moving these changes to the webpack renderer config instead?

Yeah I think we should moving these changes to renderer config, because main process may not need it.

Also I think we can have renderer base config for `development` and `production` on [`webpack.config.base.js`](https://github.com/chentsulin/electron-react-boilerplate/blob/master/webpack.config.base.js).

amilajack:
@jhen0409 Made my changes. Do they reflect what you described?

PR #602, opened 2016-12-08
Change the store to make adding middleware and enhancers more accommodating 

dustintownsend:
This is how I have my store set-up and thought it would be useful for the boilerplate. I can see this allowing for future extensibility and makes it easier to do things like creating a store for the main and renderer process (something I am experimenting with on the boilerplate).

Also what are the thoughts on making a single store for development and production?


hardchor:
Good work! I also do like the idea of a single store for dev and prod, not a big fan of the amount of code duplication that is going on in the `.development.js` files.

amilajack:
@dustintownsend would be great if you could fix the merge conflicts here. I'd love to merge this. Sorry for being so late on this 😢

Issue #108, opened 2015-12-26
[Proposal] Use redux-electron-store


jhen0409:
We can use [redux-electron-store](https://github.com/samiskin/redux-electron-store), in this way, we can let main process dispatch actions to other BrowserWindow.
1. Main process store reload: https://github.com/samiskin/redux-electron-store/issues/3
2. One issue at https://github.com/samiskin/redux-electron-store/issues/6, otherwise we can split DevTools to another BrowserWindow.

I have such a demand, but this boilerplate will become too complicated? I would like to know everyone views.

chentsulin:
This one is quite interesting. I will give it a try next month.

danielmahon:
I just started a project which would need multiple windows and would love an example on how, in this boilerplate workflow, to properly send/store state across windows...

domasx2:
I also wanted to dispatch some actions from main process, and also to share state between multiple windows.  
Achieved this quite easily by instantiating the store on the main process and then remote.requiring it in the renderer:

```javascript
const remote = require('electron').remote;
const store = remote.require('./app/store');

render(
    <Provider store={store}>
        <CounterPage/>
    </Provider>,
    document.getElementById('root')
);
```

Works perfectly. Had to hook babel into main proc though.

naderhen:
@domasx2 when you hook babel into the main proc, are you then able to successfully run `npm run package` and run the application? I receive a number of errors related to babel imports.

domasx2:
Yeah, stage-0 preset was somewhy crapping out on main proc.  
Probably the right way to solve it was to figure out what's wrong and then refac .babelrc using BABEL_ENV to set up different configs for main proc/ webpack/ tests.
But what I did was:

```javascript
//.babelrc
{}

//bootstrap.js
require('babel-register')({
    "presets": ["es2015"],
    "plugins": ["transform-object-rest-spread"],
    "ignore": /(browser\/api|common\/api)/
});
require('./main');

//package.json
...
 "main": "bootstrap.js"
...

//webpack.base.js
...
    loaders: [{
      test: /\.jsx?$/,
      loader: 'babel-loader',
      query: {
          "presets": ["es2015", "stage-0", "react"],
          "plugins": ["add-module-exports"]
      },
      exclude: /node_modules/
    }
...

//webpack.development.js
...
config.module.loaders.forEach(function(loader) {
     if (loader.loader === 'babel-loader') {
        loader.query.presets.push("react-hmre");
     }
});
...
[...]

PR #520, opened 2016-11-06
Migration to jest

amilajack:
- Proper compatibility with async/await ([mocha isnt](https://github.com/mochajs/mocha/issues/2407#issuecomment-239028717))
- [Snapshot testing](http://facebook.github.io/jest/blog/2016/07/27/jest-14.html)
- Parallelization of tests

PR #603, opened 2016-12-08
Add electron-redux module

dustintownsend:
This uses the proposed changed in PR #602. 
I have the need for syncing the state between main and renderer, so this is my first pass. Based on the discussion from Issue #108 I decided to try https://github.com/hardchor/electron-redux/ first. I may do another PR for redux-electron-store and an alternative.

@burkhardr can you look over this and see if I got any of the implementation wrong? I based it on your timesheets repo.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -4,11 +4,13 @@ import { render } from 'react-dom';
 import { Provider } from 'react-redux';
 import { Router, hashHistory } from 'react-router';
 import { syncHistoryWithStore } from 'react-router-redux';
+import { getInitialStateRenderer } from 'electron-redux';
 import routes from './routes';
 import configureStore from './store/configureStore';
 import './app.global.css';
 
-const store = configureStore();
+const initialState = getInitialStateRenderer();
+const store = configureStore(initialState, 'renderer');
 const history = syncHistoryWithStore(hashHistory, store);
 
 render(
--- a/app/index.js
+++ b/app/index.js
@@ -10,7 +10,7 @@ import configureStore from './store/configureStore';
 import './app.global.css';
 
 const initialState = getInitialStateRenderer();
-const store = configureStore(initialState, 'renderer');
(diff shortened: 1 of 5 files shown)


hardchor:
@dustintownsend Hey. Nothing glaringly obvious I can spot. 
Most importantly: does it work? 😉

hardchor:
Might be worth adding some examples of, e.g. async actions, local or aliased actions.

dustintownsend:
@hardchor do you have any insight into why the test are failing?

It is failing at replyActionMain.js at ipcMain.on. It seems that ipcMain is not importing. I assume its due to not being in electron but a test, but I don't know enough about testing to be 100%.

jhen0409:
> It is failing at replyActionMain.js at ipcMain.on. It seems that ipcMain is not importing. I assume its due to not being in electron but a test, but I don't know enough about testing to be 100%.

https://github.com/chentsulin/electron-react-boilerplate/issues/573#issuecomment-263033090 have provided a workaround, maybe we can migrate to `electron-mocha` (although we have #520), but it will broken `test-watch` (A reason is https://github.com/jprichardson/electron-mocha/pull/18).

hardchor:
If this is a unit test, you'll probably need to mock `electron`. It's slightly confusing that `electron` is now also the name of `electron-prebuilt` on npm, so whilst running tests, that'll probably be imported (rather than the actual electron helpers).

dustintownsend:
I'm going to close this for now. I'm not sure this is a good fit for the boilerplate right now. I think it would be better as an optional add on.

hardchor:
What are the core maintainer's thoughts? I could have another go at this once the jest PR is in.

/cc @jhen0409 @amilajack

amilajack:
I'm not very experienced with redux. @jhen0409 and @chentsulin may have more to say about this

PR #604, opened 2016-12-08
Fix #423, Dynamic Script Injection Warning

### LINKED ISSUE #423: Possible performance impact of dynamic script injection
when i run   nam run dev, and the app is running ,but it show error message like this

A Parser-blocking, cross-origin script, http://localhost:3000/dist/bundle.js, is invoked via document.write. This may be blocked by the browser if the device has poor network connectivity.

why? I use mac os 10.11.5, it cannot work,
and  i use window, it ok,
how can i do??

amilajack:
**DONT MERGE. NEEDS TESTING**

Inspired by https://github.com/chentsulin/electron-react-boilerplate/issues/423#issuecomment-265875725

### DIFF
--- a/app/app.html
+++ b/app/app.html
@@ -9,7 +9,7 @@
           const link = document.createElement('link');
           link.rel = 'stylesheet';
           link.href = './dist/style.css';
-          document.body.appendChild(link.outerHTML);
+          document.body.appendChild(link);
         }
       }());
     </script>
@@ -23,7 +23,7 @@
         script.src = (process.env.HOT)
           ? 'http://localhost:' + port + '/dist/bundle.js'
           : './dist/bundle.js';
-        document.body.appendChild(script.outerHTML);
+        document.body.appendChild(script);
       }
     </script>
   </body>
--- a/app/app.html
+++ b/app/app.html
@@ -9,7 +9,7 @@
           const link = document.createElement('link');
           link.rel = 'stylesheet';
           link.href = './dist/style.css';
-          document.write(link.outerHTML);
+          document.body.appendChild(link.outerHTML);
         }
       }());
     </script>
@@ -23,7 +23,7 @@
         script.src = (process.env.HOT)
           ? 'http://localhost:' + port + '/dist/bundle.js'
           : './dist/bundle.js';
-        document.write(script.outerHTML);
+        document.body.appendChild(script.outerHTML);
       }
     </script>
   </body>


dustintownsend:
So I was mistaken that it generated the same output. I didn't realize that the css was loaded in differently for development. I missed the `if (!process.env.HOT)` statement ha. And it appears that document.body.appendChild will not put something in the head. Sorry about that.

The one in the body for the script is good. But the head link will need to be changed back to use document.write. That at-least removes the warning from the console during development, but we will need to come up with an alternative to load the link if we want to eliminate the use of document.write fully.

I feel like the correct solution would be to use webpack to load style.css link in for production similar to development.

app/app.html
```diff
@@ -9,7 +9,7 @@
           const link = document.createElement('link');
           link.rel = 'stylesheet';
           link.href = './dist/style.css';
-          document.body.appendChild(link.outerHTML);
+          document.body.appendChild(link);
```
dustintownsend: change back to `document.write(link.outerHTML)` for now

app/app.html
```diff
@@ -23,7 +23,7 @@
         script.src = (process.env.HOT)
           ? 'http://localhost:' + port + '/dist/bundle.js'
           : './dist/bundle.js';
-        document.body.appendChild(script.outerHTML);
+        document.body.appendChild(script);
```
dustintownsend: This should be good.

amilajack:
> I feel like the correct solution would be to use webpack to load style.css link in for production similar to development.

I'll close this PR and create a new one with the proper webpack implementation. I'll do some research into this.

dustintownsend:
Looks like this will work: for the link tag: `document.getElementsByTagName('head')[0].appendChild(link);` I tried this and it passes all the test.

Webpack might be a cleaner solution. I tried a few things but kept running into road blocks, but that is likely from my lack of webpack understanding, fully at-least. It looks like style-loader is suppose to do this, but since we are using css-modules and the extract text plugin it is expected to just hard code the link ref in... at-least from what I could figure out. Also the HTML plugin will add the link tag but it adds it the actual source file, so would need to be removed for development.

amilajack:
@dustintownsend I'll reopen the PR. Can you make the changes directly to this PR? That would be more convenient because you know exactly how to fix it. I believe GitHub allows that now. If not, PR to this branch and I will merge the PR into this branch.

dustintownsend:
It took me a minute to figure out that your PR was a branch in this repo and not one in your account ha. But I think I got it figure out. See #608 

I ran the test on my work machine just now and everything looks good. I'll check it on a windows machine later today.

amilajack:
Builds are failing on travis because they seem to be using the wrong version of node

amilajack:
@dustintownsend re-ran the tests and all ci tests pass. I'll try this locally and let you know if everything works out correctly

amilajack:
@dustintownsend finished testing and everything seems to work. Can you confirm if the same is true on your machine?

dustintownsend:
@amilajack I'm doing a fresh clone now to test things out. Sorry I didn't check my github notifications this weekend. I need find an app so I get notifications on my phone.

dustintownsend:
@amilajack everything looks good. Just tested it out on Windows 10. 
I tested `npm run dev` `npm run build` then `npm run test` and `npm run test-e2e`. Then finally `npm run package`.

PR #606, opened 2016-12-09
Updated eslint flowtype errors

amilajack:


amilajack:
screen shot 2016-12-08 at 8 42 10 pm

React prop support ❤️

PR #607, opened 2016-12-09
Flowtype migration

amilajack:
**⚠️ Do Not Merge**

This essentially requires the user of the boilerplate to know at least a little flow. Lets merge this when we provide some learning resources and documentation on the tools that this boilerplate uses

- [x] Update `eslint-plugin-flowtype-errors`
- [x] Add `eslint-plugin-flowtype`
- [x] Add `babel-plugin-tcomb`
- [x] Migrate react `static propTypes` to flow types


amilajack:
@danielbushman for some reason I can add you as a reviewer. Would be great if you could review this.

When should this be merged? The `propTypes` examples in the boilerplate use flow: https://github.com/chentsulin/electron-react-boilerplate/blob/flowtype-props-migration/app/components/Counter.js#L7-L13

This allows for better static analysis when developing. Also if I'm not mistaken, `propTypes` will be deprecated at some point.

Is this okay to be merged now?

amilajack:
screen shot 2016-12-08 at 8 42 10 pm

amilajack:
@dustintownsend accidentally tagged the wrong person. Meant to tag you in https://github.com/chentsulin/electron-react-boilerplate/pull/607#issuecomment-266163847. Would be great if you could respond to it thanks.

dustintownsend:
@amilajack I really need to learn more about flow. I assume it's basically just making javascript more like typescript. For the benefit of catching typos/code mistakes in the editor.

I'm not 100% sure what I'm looking for here, but I see the same warnings from #610 but in the editor. ESLint warnings for `flowtype/no-weak-types`. I assume this is what is expected and if so it's working. 
I am running Atom with the recommended packages from the readme.

amilajack:
@dustintownsend that's a linter error that we get from using very weak/generic types like so:

```js
function foo(x: Object, y: Function) {}
```

`Object` and `Function` are weak types. Its better to do something like so:

```js
type x = {
  name: string,
  age: number
}
function foo(x: x, y: () => void) {}
```

amilajack:
@dustintownsend Hmm try restarting the editor. 

Here's what I see:

screen shot 2016-12-12 at 8 41 19 am

Try restarting atom

dustintownsend:
@amilajack if I add `/* eslint flowtype/no-weak-types: 2 */` to the file I get the `Error` instead of `Warning`.

Is that the expected behavior?

PR #610, opened 2016-12-11
added lintfix script in package.json

ritz078:


amilajack:
@ritz078 would be great if you could test this locally just in case. I've done so and it works for me

jhen0409:
@amilajack would be better if you can rebase commits appropriately, in the past I've seen a lot of commits may not be necessary on `master`.

If we need to check for past changes, I think it will be bad experience for me. :\

amilajack:
@jhen0409 Sorry about that. I didn't look at the history in a while and assumed GitHub did this. From now on, I will rebase and squash. What is the recommended way of doing this? Is there a way that Github provides?

dustintownsend:
@amilajack I'm not 100% sure what I'm looking for..

I ran `npm run lint-fix` and I get the following result...

  app\actions\counter.js:18:21
  ‼  18:21  Unexpected use of weak type "Function"  flowtype/no-weak-types
  ‼  18:41  Unexpected use of weak type "Function"  flowtype/no-weak-types
  ‼  30:21  Unexpected use of weak type "Function"  flowtype/no-weak-types

  app\reducers\counter.js:4:60
  ‼   4:60  Unexpected use of weak type "Object"    flowtype/no-weak-types

  app\store\configureStore.development.js:35:54
  ‼  35:54  Unexpected use of weak type "Object"    flowtype/no-weak-types

  app\store\configureStore.production.js:12:54
  ‼  12:54  Unexpected use of weak type "Object"    flowtype/no-weak-types

  6 warnings

I am testing on Windows 10 today with a fresh clone of master (since this was already merged in).

amilajack:
@dustintownsend that's a linter error that we get from using very weak/generic types like so:

```js
function foo(x: Object, y: Function) {}
```

`Object` and `Function` are weak types. Its better to do something like so:

```js
type x = {
  name: string,
  age: number
}
function foo(x: x, y: () => void) {}
```

jhen0409:
> I didn't look at the history in a while and assumed GitHub did this. From now on, I will rebase and squash. What is the recommended way of doing this? Is there a way that Github provides?

@amilajack, GitHub provided `Squash and merge` for PR, It is appropriate for this PR I think. In my habits for part of commits, I using `git rebase --interactive <id>` to fix some commits (and `--continue`), or `git reset` to remove wrong commit rather than directly add new ones on a PR.

PR #618, opened 2016-12-14
add build productName

TimoSolo:
The build productName is used for the app distributables... suggest to add it here so other people can easily change it without digging through the docs!


amilajack:
🎉

PR #622, opened 2016-12-15
Build Optimizations

amilajack:


amilajack:
@dustintownsend for some reason, I cant add you as a reviewer:

screen shot 2016-12-14 at 10 48 30 pm

Guess I have to comment here to let you know to review this. thanks!

amilajack:
LOL i just noticed that GitHub now has both contributors and collaborators. @chentsulin would be great if you could add @dustintownsend as a collaborator. He's done a lot for this project.

.babelrc
```diff
@@ -1,5 +1,9 @@
 {
-  "presets": ["es2015", "stage-0", "react"],
+	"presets": [
+    ["env", {"targets": { "electron": 1.4 }, "useBuiltIns": true }],
```
jhen0409: I think we should use `"node": 6` target, we need use for `build`, `test` and `hot-server`, make sure that node and electron will be different in the future, although currently they are the same. @amilajack What do you think?

dustintownsend:
The babili-webpack-plugin works good. I just updated my project to use it as a test and it removed all those ugilifyjs warnings/errors I was getting!

.babelrc
amilajack: @jhen0409 makes sense. At the moment, I believe node 6 and 7 have the exact same es6 compatibility.

chentsulin:
Ok, I Already sent collaborator invitation to @dustintownsend.

amilajack:
@jhen0409 Made all the changes you requested. Anything else?

PR #634, opened 2016-12-24
Remove api dir

amilajack:


amilajack:
Not sure why this api dir is here. Does anyone know?

rahilsondhi:
I was using it for classes that talk to the main electron process. If they don't belong in `api/`, then where?

amilajack:
@rahilsondhi the `./api` dir is project specific. That was the incentive for removing it

PR #642, opened 2016-12-30
Simplified dev script in readme

amilajack:
I think we should simplify the readme the development instructions. I think most users will not need the 'run scripts'. Most probably use `npm run dev`. Thoughts?


jhen0409:
It's a quick way to restart main process, avoid wasting time restart webpack server. I still using both commands, and think it's good to keep it.

Maybe we can change the order of precedence?

amilajack:
Good idea. I'll update the PR to reflect this

amilajack:
I have a few question: what do `hot-server` and `start-hot` do?

PR #656, opened 2017-01-11
Adding missing dev dependencies

zkochan:
babel-plugin-transform-class-properties and babel-plugin-transform-es2015-classes are specified in `.babelrc` but were not in dev dependencies

I assume this works fine with npm@3 and yarn, because they have flat dependency trees. However, these packages should be in dev deps if they are used.

I noticed this when I was checking a really old issue created at pnpm https://github.com/rstacruz/pnpm/issues/79. By the way, I have good news! I worked the last 2 months on a complete rewrite of pnpm and with the latest pnpm@0.48.0, electron-react-boilerplate works with no issues.

This is the only issue, that these 2 deps are not in dev deps

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,8 @@
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-dev-expression": "^0.2.1",
     "babel-plugin-tcomb": "^0.3.24",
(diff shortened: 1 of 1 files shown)


jhen0409:
Thank you @zkochan!

PR #674, opened 2017-01-24
Revert "Update babili-webpack-plugin to version 0.0.8 🚀"

amilajack:
Reverts chentsulin/electron-react-boilerplate#670

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "babel-preset-react-optimize": "^1.0.1",
     "babel-preset-stage-0": "^6.16.0",
     "babel-register": "^6.18.0",
-    "babili-webpack-plugin": "^0.0.8",
+    "babili-webpack-plugin": "^0.0.7",
     "boiler-room-custodian": "^0.6.1",
     "chai": "^3.5.0",
     "concurrently": "^3.1.0",


jhen0409:
I realized that problem might be from `babel-plugin-minify-dead-code-elimination@0.1.2` (`0.1.1` works fine), so it will still not work even if we reverted the webpack plugin version.

amilajack:
What about temporarily disabling babli?

amilajack:
Merging because of https://github.com/chentsulin/electron-react-boilerplate/issues/672#issuecomment-274996314. I also confirmed this with my own local revert

jhen0409:
We must ensure CI is passing.. I guess somebody are still use `babel-plugin-minify-dead-code-elimination@0.1.1` because yarn lock, we can get `0.1.2` by npm, or `yarn upgrade`.

I'll submit a better temporarily fix for that.

amilajack:
This seems to have fixed it. See https://travis-ci.org/chentsulin/electron-react-boilerplate/builds/195083218, which is a test of a branch that has this change. It seemed like the cache was outdated

jhen0409:
@amilajack looks like this passed build is still used 0.1.1. (see [this line](https://github.com/chentsulin/electron-react-boilerplate/blob/e46f10d5bda20761f27547df355541b9db81e90d/yarn.lock#L593))

Issue #725, opened 2017-02-10
Add .nvmrc file


cmbirk:
I'm having issues running `npm install`, and it has to do with a differing node version.  I can open a PR for this, but wanted to log the ticket first.

amilajack:
🎉🎉🎉 PR away!

cmbirk:
Cool.  I'll add it as `v6.9.5`.  PR coming soon

PR #730, opened 2017-02-11
add .nvmrc file

cmbirk:
This just adds the last release for node 6.* in the `.nvmrc` file


cmbirk:
Fixes #725

amilajack:
Is there a way to allow nvm to use any version >= 6.0.0?

cmbirk:
Not that I'm aware of.  The idea is to specify the latest node version that this supports and remove ambiguity.

amilajack:
Thanks!

PR #763, opened 2017-02-17
Switch to single entry point for electron

akozhemiakin:
Electron has a known problem that it ignores package.json file when it is run with explicitly specified entry point (and after looking through some electron project issues and questions, after couple conversations on their Slack channel, I believe they consider it not as a bug but as a feature). It leads to some negative consequences like for example wrong app name and appData and userData directories (If electron ignores package.json it sets app name to 'Electron').
Currently we have this script to run in the hot mode: `"start-hot": "cross-env HOT=1 NODE_ENV=development electron -r babel-register -r babel-polyfill ./app/main.development"` which leads to aforementioned problems.
This pull request creates single entry point and let electron use package.json meta data in both dev and prod modes.

Now, before merging this, I ask you guys to test it in various scenarios (dev\prod run\build etc) because I'm not sure that I covered them all.


amilajack:
@chentsulin @dustintownsend Thoughts on this?

dustintownsend:
I do agree this could be an issue. Especially for someone that is working on more than more erb based project. Personally I solve the issue of the wrong directory being used by calling `electron.app.setPath('appData/userData', correctPath)`. We could parse the `app/package.json` and set the paths for development.

What if we call `webpack --config webpack.config.electron.js` before calling the `start-hot` script?

My only concern with this pull request is adding the `root.js` file. But it does solve a problem, so if others are okay with it I think it is fine. Maybe we could call it `main.entry.js` so it could be more descriptive than root.

akozhemiakin:
> Personally I solve the issue of the wrong directory being used by calling electron.app.setPath('appData/userData', correctPath). We could parse
the app/package.json and set the paths for development.

I thought about such workarounds, but I afraid that electron have or eventually will have some other stuff dependent on meta data from package.json. So I think that instead of duplicating electron logic with necessity to monitor any changes on that front and adjust erb each time, the more reliable and consistent approach will be to provide electron with a meta data it needs to play well both in dev and prod environments.

Yeah, I'm a bit unsure about the file name too. main.entry.js sounds well.

akozhemiakin:
@dustintownsend done

amilajack:
Will merge this over the weekend after manual testing

akozhemiakin:
@amilajack Did you close this without merging intentionally or by mistake?)

amilajack:
Mistake. My bad.

akozhemiakin:
I just resolved merge conflicts with a latest master

akozhemiakin:
Can we merge it or should I change anything else here? It would be awesome if we can merge it before any other conflicting changes to master :)

amilajack:
Is it possible to add these changes without creating new files? A common issue people have with this boilerplate is that they dont understand the configuration and the responsibility of each file/module. I feel like adding more configuration files will only make this issue worse. Is there a way to bring these changes without creating new files (namely `main.entry.js`)?

akozhemiakin:
Now we have two different target files for electron and select one of them explicitly in each npm script. To eliminate the problem this PR is trying to solve, we need to provide electron with a single entry point (I do not know any other way how it may be solved and, actually, this is what one of the electron contributors suggested to me). Correct me if I'm wrong, but I doubt that we can merge different (dev and prod) scripts that we have for electron right now. So, the only option left is to create another file that will serve as a central entry point.

talha131:
@amilajack any update on status of this PR? Thank you.

akozhemiakin:
I will be ready to resolve merge conflicts as soon as we come to some conclusion about the content of this PR.

alexcroox:
Electron Vue recently switched to a single package.json, I wonder if that would help with any inspiration: https://github.com/SimulatedGREG/electron-vue

PR #613, opened 2016-12-12
Migrate to jest

amilajack:

Issue #504, opened 2016-11-01
Roadmap for Release 1.0.0


amilajack:
## Release 0.11.0
- [x] Fix `debugger` statement devtools integration
- [x] `Babli` and `babel-preset-env` support
- [x] Migration to Jest
- [x] Webpack DLL support
- [x] Document configuration
- [x] react-router-4 migration
- [x] Route based code splitting
- [x] webpack-bundle-analyzer

## Release 0.11.1
- [x] Support debugging packaged app

## Release 0.12.0
- [x] Folder refactoring, add `./internals` dir

## Release 1.0.0
- [x] Complementary website
- [x] Auto Updater
- [x] Artifact support
- [x] Add `electron-redux` support
- [ ] Debugging main/renderer processes
- [ ] Video Tutorial Series

Please comment if you want to discuss the addition/removal of features.

ChinW:
Hot reload main process would be cool and useful for me.

dustintownsend:
What about adding some of these:
- Redux Saga (instead of thunk)
- Redux Persist
- Redux Sauce (to reduce Redux boilerplate)
- ImmutableJs or Seamless Immutable 
- Reactotron (not sure if it would provide more than current dev tools or not)

Not sure if those types of things would be out of scope for this boilerplate or not?

amilajack:
I think we should use Reactotron and redux persist. I'm not very familiar with redux and its ecosystem so I'm not sure about redux-saga. What does it do? Does it require the user to know about it for them to use the app?

tsemerad:
redux-saga is a good solution for handling async logic, or any other side-effects (such as dispatching redux actions in response to other actions). It is often used instead of redux-thunk, which is also used for the same purposes. I prefer redux-saga to redux-thunk because it provides many high-level tools and patterns for managing async logic, but it does come at a significant learning overhead compared to redux-thunk.

> Does it require the user to know about it for them to use the boilerplate?

It is definitely significantly more complex than using redux-thunk. It requires learning about generator functions and redux-saga's "effects" system, which has great documentation, but is significant.

As for whether it should be included in the boilerplate, that's a good question. I'm on the fence.

rsaccon:
[...]

PR #768, opened 2017-02-21
Switch mocha to jest

akozhemiakin:
This commit replaces mocha with jest. Also this commit removes dependency on
babel-plugin-webpack-loaders plugin because of the following reasons:
1. It seems that the author deprecated and abandoned this plugin and he suggests
not to use it and to use mocks and native test utilities tools instead.
2. This plugin conflicts with webpack 2 and because of the reason one it is unlikely
that it will ever support it.

In order to remove dependency on the aforementioned plugin and keep using css
modules in the same time the following trick is used: All components that need to
be selected in the end-to-end test receive data-tid (TestID) and data-tclass
(TestCLASS) properties. It is necessary because without running webpack we can't
use class names to select components because we do not know what class names
css loader actually generates for us.

P.S. I prepared upgrade to webpack 2 but it depends on this commit. So after this PR is closed I'll create pull request for webpack upgrade.


jhen0409:
Could you also change `test/example.js`? Has [lint errors in AppVeyor build](https://ci.appveyor.com/project/chentsulin/electron-react-boilerplate/build/1406/job/c1siw00d33rsu85a#L1692), I guess Travis CI passed because `node_modules` cache.

akozhemiakin:
I fixed example.js but it seems that appveyor has some other reason to fail. I got very similar error one time and fixed it by deleting node_modules directory and reinstalling project. Do not know however how to fix appveyor.

jhen0409:
Actually @amilajack was do #613, we have the same problem on AppVeyor. :\

akozhemiakin:
Maybe it expects \ instead of / ? Apveyor runs on windows after all.

akozhemiakin:
Hooray! I fixed it )

akozhemiakin:
Done. I believe I remove all references to mocha.

amilajack:
@akozhemiakin LGTM!

akozhemiakin:
Nice. Know @amilajack you can fill another checkbox here #504 (:

amilajack:
Done!

PR #769, opened 2017-02-22
Webpack2 migration

AlexFrazer:
Just makes webpack 2 work.

Guides here: https://webpack.js.org/

Adds a few dependencies (`react-hot-loader`), removes a few dependencies as well (`webpack-validate`).

The biggest change, however, is that there is no longer a `server.js` file. Seemed pretty superfluous. Easier to just spawn it in the devServer setup.


webpack.config.development.js
```diff
@@ -79,11 +121,28 @@ export default validate(merge(baseConfig, {
...
+  target: 'electron-renderer',
+  devServer: {
+    port,
+    hot: true,
+    inline: false,
+    historyApiFallback: true,
+    contentBase: path.join(__dirname, 'dist'),
+    publicPath,
+    setup() {
+      spawn('npm', ['run', 'start-hot'], { shell: true, env: process.env, stdio: 'inherit' })
```
jhen0409: We need to avoid it run with `npm run hot-server`, that's why we have a `--start-hot` flag in `server.js`.

AlexFrazer: I'm sorry, not sure what you mean by that... do you want to splice the `process.env.argv`, check for the `start-hot` flag, and conditionally spawn the child process?

jhen0409: Ah looks like you haven't change `hot-server` script, so it will broken. You may need to move the current `dev` to `hot-server` in npm scripts.

I expected it can [`run these two commands simultaneously in different console tabs`](https://github.com/chentsulin/electron-react-boilerplate#run), it will be different from `npm run dev`.

> do you want to splice the process.env.argv, check for the start-hot flag, and conditionally spawn the child process?

Yeah.

AlexFrazer: I don't think that will work. If you try passing `npm run hot-server -- --start-hot`, then `webpack-dev-server` will reject it as an invalid argument.

One alternative option is to export a function and pass it in as the `env`, eg:

```
export default function devConfig(env = 'start-hot') {
  return merge(baseConfig, {
    // ...
    devServer: {
      // ...
      setup() {
        if (env === 'start-hot') {
          spawn('npm', ['run', 'start-hot'], { shell: true, env: process.env, stdio: 'inherit' })
            .on('close', code => process.exit(code))
            .on('error', spawnError => console.error(spawnError));
        }
      }
    }
  });
}
```
Thoughts?

jhen0409: Maybe just `"dev": "cross-env START_HOT=1 npm run hot-server"` and check `process.env.START_HOT` will be cleaner.

AlexFrazer: Gotcha -- just pushed it

package.json
```diff
@@ -11,7 +11,7 @@
     "test-e2e": "cross-env NODE_ENV=test BABEL_DISABLE_CACHE=1 mocha --retries 2 --compilers js:babel-register --require ./test/setup.js ./test/e2e.js",
     "lint": "eslint --cache --format=node_modules/eslint-formatter-pretty .",
     "lint-fix": "npm run lint -- --fix",
-    "hot-server": "cross-env NODE_ENV=development node --max_old_space_size=2096 -r babel-register server.js",
+    "hot-server": "cross-env NODE_ENV=development node -r babel-register node_modules/.bin/webpack-dev-server --config ./webpack.config.development.js",
```
dustintownsend: Changing the path to `./node_modules/webpack-dev-server/bin/webpack-dev-server` fixed the issue I was running into on Windows.

I also think we should add `./` in-front of node_modules, so it matches the syntax of the other scripts, but it does work with it out for me.

webpack.config.development.js
```diff
@@ -3,71 +3,113 @@
  * Build config for development process that uses Hot-Module-Replacement
  * https://webpack.github.io/docs/hot-module-replacement-with-webpack.html
```
dustintownsend: Remove the link to webpack v1 doc or replace with v2 doc
https://webpack.js.org/concepts/hot-module-replacement/

.babelrc
```diff
@@ -1,6 +1,7 @@
 {
   "presets": [
     ["env", { "targets": { "node": 6 }, "useBuiltIns": true }],
+    ["es2015", { "modules": false }],
```
jhen0409: Could you explain why it added back?

AlexFrazer:
Hey guys, sorry for the long arduous process in the review. Commit history is a bit of a mess

dustintownsend:
@AlexFrazer it appears that `extract-text-webpack-plugin` no longer has a `beta` version, so that needs to be set to `2.0.0` in the `package.json` or you can't get past the initial `npm install`.

Other than that it appears everything is working good on Windows!

One question I have is what is the reason for adding the `Root` container component in addition to the `App` container component?

dustintownsend:
@AlexFrazer this is odd. When I do `npm run lint` I am getting a `linebreak-style` for `app\containers\CounterPage.js`. It appears to be in `CRLF` instead of `LF`. But, it doesn't appear you made any changes to that file. I changed the line style in Atom and ran the lint again and it passed. Maybe something on my end.

dustintownsend:
@AlexFrazer based on this https://webpack.js.org/guides/migrating/#occurrenceorderplugin-is-now-on-by-default we can remove the init of the `OccurrenceOrderPlugin` in `webpack.config.production.js`

AlexFrazer:
@dustintownsend , The root container is just hot module replacement -- I put it in a separate module so it could ideally be conditionally imported. I've also run into some strange errors when it wasn't a separate module, not sure what the deal with those were.

Addressed the two other comments on versions and OccurenceOrderPlugin

dustintownsend:
@AlexFrazer I migrated the changes (minus the Root component) you made here to my project at work and didn't noticed any issues! I will have to test the HMR more and see if I need to apply that change as well. 

I will do more testing on this as soon as I can.

Good work!

dustintownsend:
@AlexFrazer HOT was broken in my project. When I added the Root component it worked!

I've tested this PR on windows a few times now over the past couple days and everything is working good for me. 

@jhen0409 was there anything outstanding from your review? 

@amilajack is there anything outstanding on this PR? Does anything need to be done for Flow types in this PR?

Maybe we can get this merged!

amilajack:
I want to manually test this on mac just to make sure everything works. Will probably do this over the weekend

amilajack:
>  is there anything outstanding on this PR? Does anything need to be done for Flow types in this PR?

No. Flow and webpack are not not coupled in this boilerplate. babel and flow are. There shouldn't be an issue

amilajack:
@AlexFrazer Thanks for taking the time to make this PR!

AlexFrazer:
@amilajack @dustintownsend @jhen0409 Thanks for taking the time to review it, as well

PR #774, opened 2017-02-24
Fix and a little refactoring for testing process

akozhemiakin:
I was so focused on fixing jest execution on windows (to make appveyor happy) and (relying on travis) totally forgot to manually test it on linux. So I was very surprised after "npm run test" execution failure in my Ubuntu. It seems that travis executes it in some environment that allows it to run successfully. After short (not so short actually) investigation I came to a conclusion that the problem is in the differences between linux and windows (and their various execution environments) escaping strategies and requirements. So I moved all that starting logic into separate script and used cross-spawn (https://github.com/IndigoUnited/node-cross-spawn) to run it smoothly.

Also, as a bonus, I moved end-to-end test to separate folder and change end-to-end tests selection strategy to run all tests from this and nested folders. So now it is possible to have multiple end-to-end tests which must be placed in ./test/e2e/... in order to be recognized as end-to-end tests.


amilajack:
@akozhemiakin Thanks for the PR! 🎉 🎉 🎉

Issue #773, opened 2017-02-23
stage-0 seems unstable. Move it down to stage-2?


AlexFrazer:

amilajack:
Agreed. PR would be welcome!

luigiplr:
I would argue that `stage-3` would be a safer bet.

In `stage-2` the semantics of a proposed feature may still change (e.g. decorators). Whereas in `stage-3` you can feel more secure knowing that the spec is complete & will be integrated in all browsers unless a critical issue comes up.

Further Reading:

- https://tc39.github.io/process-document/
- https://babeljs.io/docs/plugins/preset-stage-3/
- https://babeljs.io/docs/plugins/preset-stage-2/

amilajack:
I agree with stage-3

amilajack:
Closing as the PR for this is in progress

PR #775, opened 2017-02-24
Change babel traspilation options

KELiON:
* Use electron version instead of node version for `babel-preset-env`
* Use `stage-2` instead of `stage-0` (#773)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [
-    ["env", { "targets": { "electron": "1.4" }, "useBuiltIns": true }],
+    ["env", { "targets": { "node": "current" }, "useBuiltIns": true }],
     "stage-2",
     "react"
   ],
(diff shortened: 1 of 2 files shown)


jhen0409:
The target `"node": 6` was discussed in https://github.com/chentsulin/electron-react-boilerplate/pull/622#discussion_r92592708, and these two are only use `es2015-module-commonjs` for currently.

If we want to use `electron` target for future changes, I'm recommend to write an another babel config in `webpack.config.base.js` for Electron runtime usage.

amilajack:
@KELiON Would be great if you could use babel-preset-env's ['current'](https://github.com/babel/babel-preset-env#support-a-target-option-node-current-to-compile-for-the-currently-running-node-version) option 😄

amilajack:
@jhen0409 Fixed this

jhen0409:
Currently [babel-preset-env](https://github.com/babel/babel-preset-env) should included `latest` preset, so it may remove `transform-async-to-generator` if current node is `^7.6`, I think that will broken electron 1.4 if we use async function.

Maybe also add `electron: 1.4`?

amilajack:
Let's wait for `electron: "current"`

https://github.com/babel/babel-preset-env/pull/191

PR #779, opened 2017-02-27
React toolbox

ballPointPenguin:
a few things are just cosmetic, like line spacing in the webpack configs,
also the package versions probably don't matter here.
the "browserslist" and "category" attributes of the package.json were giving me trouble building, especially on linux and don't seem to be required.
The main things are the inclusion of some additional packages, the postcss.config, and removing the extract-text stuff from the production webpack config.


ballPointPenguin:
sorry I meant this PR to be to my forked repo only; closing.

PR #791, opened 2017-03-05
Refactoring code for menu for easier changes

axsauze:
When modifying the code for the menu I thought it would be a good idea to separate it from the main.development.file, add it to a class and break it down to make it easier for people to add changes  and customise menus in the future.

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -1,3 +1,4 @@
+// @flow
 import { app, BrowserWindow } from 'electron';
 import MenuBuilder from './menu';
 
@@ -52,8 +53,8 @@ app.on('ready', async () => {
   mainWindow.loadURL(`file://${__dirname}/app.html`);
 
   mainWindow.webContents.on('did-finish-load', () => {
-    mainWindow.show();
-    mainWindow.focus();
+    mainWindow.show(); // eslint-disable-line
+    mainWindow.focus(); // eslint-disable-line
   });
 
   mainWindow.on('closed', () => {
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -53,8 +53,11 @@ app.on('ready', async () => {
   mainWindow.loadURL(`file://${__dirname}/app.html`);
 
   mainWindow.webContents.on('did-finish-load', () => {
-    mainWindow.show(); // eslint-disable-line
-    mainWindow.focus(); // eslint-disable-line
+    if (!mainWindow) {
+      throw new Error('"mainWindow" is not defined');
+    }
+    mainWindow.show();
+    mainWindow.focus();
   });
 
   mainWindow.on('closed', () => {
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -4,8 +4,6 @@ import MenuBuilder from './menu';
 let menu;
 let mainWindow = null;
 
-const urlDriveHome = 'https://drive.google.com';
-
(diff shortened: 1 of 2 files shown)


axsauze:
All checks are now passing

jhen0409:
Use `env` as an argument will breaks deadcode elimination for webpack, I think keep using `if (process.env.NODE_ENV === 'development') { ...statements }` will be better.

axsauze:
Thanks for the heads up @jhen0409 - I've updated it now

axsauze:
@jhen0409 just following up - what is the process to confirm whether this pull request is relevant or not?

amilajack:
> process to confirm whether this pull request is relevant or not?

This PR is relevant. Thanks for it! 😄

axsauze:
Perfecto! Thanks @amilajack :) Let me know if any other changes required, happy to re-refactor as required.

axsauze:
Perfect @amilajack - I have added the // @flow on top of the file.

Is there a reason why **main.development.js** doesn't have the // @flow pragma at the top of the file?

amilajack:
@axsauze I'm not sure. But i would recommend adding it if possible.

axsauze:
Perfecto - I've added the flow pragma at the top of the file now, and made the relevant changes in the file.

axsauze:
Perfecto @amilajack - I've added the flow pragma at the top of the file now, and made the relevant changes in the file.

app/main.development.js
```diff
@@ -52,8 +53,8 @@ app.on('ready', async () => {
   mainWindow.loadURL(`file://${__dirname}/app.html`);
 
   mainWindow.webContents.on('did-finish-load', () => {
-    mainWindow.show();
-    mainWindow.focus();
+    mainWindow.show(); // eslint-disable-line
```
amilajack: What is the error that is being disabled here?

axsauze: This is the error that appears:

``` bash

  app/main.development.js:56:6
  ✖  56:6  call of method show:  Method cannot be called on possibly null value null. See line 56   flowtype-errors/show-errors
  ✖  57:6  call of method focus:  Method cannot be called on possibly null value null. See line 57  flowtype-errors/show-errors

```

axsauze: The solution would be to add code that ensures the value of ```mainWindow``` is not null, something like:

```

  mainWindow.webContents.on('did-finish-load', () => {
    if (mainWindow === null) { mainWindow = { show: () => {}, focus: () => {} } } 
    mainWindow.show(); 
    mainWindow.focus(); 
  });
```

axsauze: But I think it would not be worth adding that code just to circumvent the eslint error in this case - what do you think @amilajack ?

amilajack: Try this:

```js
mainWindow.webContents.on('did-finish-load', () => {
  if (!mainWindow) {
    throw new Error('"mainWindow" is not defined')
  }
  // ...
});
```

axsauze: Interesting - that is much cleaner! I have updated this in the latest commit.

amilajack:
LGTM! I just want to manually test this out before merging.

axsauze:
Sounds perfect - do let me know if there are any more changes required

amilajack:
@axsauze tested locally and everything looks fine! 🎉 🎉 🎉  Thanks for this!

Issue #812, opened 2017-03-15
native modules packed to the wrong location


chilijung:
Hi, I'm trying to use this native module https://github.com/santigimeno/node-pcsclite, in my electron app. Everything works well in dev mode.

after

```
$ npm run package-all
```
and installed the `.dmg` in my mac, I couldn't open the app 

![screen shot 2017-03-15 at 11 57 05 am](https://cloud.githubusercontent.com/assets/1216029/23933029/8a261176-0976-11e7-99d5-2396deedddd0.png)

and I've try to navigate to the directory and I got this

```
➜  app pwd
/Applications/Jozing.app/Contents/Resources/app
➜  app ls
app.html     main.js      package.json
dist         main.js.map
```

in `/Applications/Jozing.app/Contents/Resources/app` I couldn't find directory `build`, Is there a way I could setup where my gyp `.node` goes? Thanks.

amilajack:
Are you using node-loader?

chilijung:
nope

amilajack:
That is probably why. See this related issue: https://github.com/chentsulin/electron-react-boilerplate/issues/807

chilijung:
oh, I see. Actually I'm using a wrapper on top of it https://github.com/tomkp/smartcard . So if I'm going to use `node-loader` to solve. I'll have to modified both https://github.com/tomkp/smartcard and https://github.com/santigimeno/node-pcsclite to require binary file via `node-loader`.

I'm thinking is there a way I could built my `.node` file to `/Applications/Jozing.app/Contents/Resources/app/build` while packaging the app.  Is this possible? And I'm curious where is `/Applications/Jozing.app/Contents/Resources/app/build` this setting be set? Thanks for quick reply!

amilajack:
> I'll have to modified both https://github.com/tomkp/smartcard and https://github.com/santigimeno/node-pcsclite to require binary file via node-loader.

I don't think you have to modify these files. I believe the way that it works is if it sees that your library code as something like `require('some-file.node')`, node-loader will uses some special kind of parsing for that specific extension and this will allow it to be bundled in your app. Try adding the web pack config for `node-loader` in your`webpack.config.base.js`

chilijung:
hmm, I've tried still getting the same error. Seems like it still finding `/Applications/Jozing.app/Contents/Resources/app/build/pcsclite.node`

here is my settings in `webpack.config.base.js`

```js
/**
 * Base webpack config used across other specific configs
[...]

Issue #807, opened 2017-03-13
serialport does not work in production.


wangyueming:
I require serialport in main process, it works in development environment. But it does not work after I package it with "npm run package" on mac.

I attach a picture here. Please look at it.

I checked the app.asar. The serialport.node really does not exist. Can Anybody help? Thanks in advance. 

[
<img width="418" alt="2017-03-13 10 27 34" src="https://cloud.githubusercontent.com/assets/10303419/23858703/494d41aa-083c-11e7-98e6-8fc180f832cd.png">
](url)

amilajack:
Try using [node-loader](https://github.com/webpack-contrib/node-loader)

wangyueming:
I installed node-loader and added lines below to webpack.config.base.js:
```
    }, {
      test: /\.node$/,
      use: 'node-loader'
    }]
   ...
    extensions: ['.js', '.jsx', '.json', '.node'],

```

In development environment:
If I import it in this way:
`import SerialPort from 'serialport.node';
`
There will be an error below:
```
App threw an error during load
Error: Cannot find module 'serialport.node'
    at Module._resolveFilename (module.js:455:15)

```

If I import it in this way:
`import SerialPort from '../../node_modules/serialport/build/Release/serialport.node';
`
There will be an error when I do "new SerialPort"
![2017-03-14 10 15 54](https://cloud.githubusercontent.com/assets/10303419/23904740/ebc6a6a2-0903-11e7-951e-3b6faa35277b.png)

In both cases after I package it with "npm run package" on mac and checked the app.asar, the serialport.node still does not exist, and the program can not start.

dustintownsend:
Try to import like this... `import { SerialPort } from....`

wangyueming:
If I use `import SerialPort from '../../node_modules/serialport/build/Release/serialport.node';`, I can do `SerialPort.list()`, I can not do `new SerialPort()`.

If I user `import { SerialPort } from '../../node_modules/serialport/build/Release/serialport.node';`, I even can not do `SerialPort.list()`, the error is shown below
![2017-03-15 10 16 17](https://cloud.githubusercontent.com/assets/10303419/23930672/c93eda68-0968-11e7-94f6-bb07afc49583.png)

amilajack:
I believe the library you are using requires the `serialport.node` file so you dont have to do this in your own code. You should be able to do something like
```js
import serialPort from 'serialport`
```
[...]

PR #814, opened 2017-03-15
Added node-loader suppport

amilajack:
Closes https://github.com/chentsulin/electron-react-boilerplate/issues/813

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,6 +163,7 @@
     "jest": "^18.1.0",
     "jsdom": "^9.9.1",
     "minimist": "^1.2.0",
+    "node-loader": "^0.6.0",
     "react-addons-test-utils": "^15.4.2",
     "react-test-renderer": "^15.4.2",
     "redux-logger": "^2.7.4",
--- a/package.json
+++ b/package.json
@@ -162,6 +162,7 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^18.1.0",
     "jsdom": "^9.9.1",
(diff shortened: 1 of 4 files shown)


amilajack:
I think we should also add `json-loader` while we're at this. Its pretty common as well

jhen0409:
I think this will still not work for #812 or #807, [like `serialport` use `bindings` to load `node` file](https://github.com/EmergingTechnologyAdvisors/node-serialport/blob/master/lib/bindings/darwin.js#L2) (and auto detect platform), I'm recommended just add to `app/package.json`.

> I think we should also add json-loader while we're at this. Its pretty common as well

`json-loader` is work on webpack 2 by [default](https://github.com/webpack-contrib/json-loader#json-loader), this is why [#769](https://github.com/chentsulin/electron-react-boilerplate/commit/830e2e876c780b58880680613e112679ab40c3de#diff-d9d75cc1dda9c5e912390983ac6b263aL17) removed it.

Slapbox:
Too funny. I just asked about this in Gitter. This seems to happen often.

Slapbox:
I get the following when I run that `webpack.config.base.js`

```
[1] "use" is not allowed
[2] "value" must contain at least one of [loaders, loader]
```

amilajack:
You're probably using webpack 1. Try pulling the upstream if you set it to this boilerplate

Slapbox:
Ah nevermind I didn't realize we had updated to Webpack 2 now. I will have to do some work to catch up.

PR #819, opened 2017-03-16
Prevent webpack.DefinePlugin from overriding process.env object

vinogradov-m:
Currently when production app is running, in the main process just one environment variable exists  - `NODE_ENV`. This change should prevent process.env object from being reset.

### DIFF
--- a/webpack.config.electron.js
+++ b/webpack.config.electron.js
@@ -34,9 +34,7 @@ export default merge(baseConfig, {
      * development checks
      */
     new webpack.DefinePlugin({
-      'process.env': {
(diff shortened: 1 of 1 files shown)


amilajack:
Can you explain why this code ensures that the env variable won't be changed?

vinogradov-m:
It follows from the code of DefinePlugin. When some object passed there it will be transformed to JSON string which will be evaluated further. And original value will be replaced with the result of this evaluation. https://github.com/webpack/webpack/blob/master/lib/DefinePlugin.js#L27

At the same time, when you define some value of primitive type it will be added to existing object (or this parent object will be created).

PR #841, opened 2017-03-21
Migrate to react-router v4 and code-splitting

Elyx0:
As mentioned in https://github.com/chentsulin/electron-react-boilerplate/issues/808
This PR switches to React router v4 with dynamic import for code splitting.

I first tried using the `bundle-loader` but then went toward the `import` promise as specified from https://reacttraining.com/react-router/web/guides/code-splitting and didn't use `System.import` as it will deprecate in Webpack3 ( https://survivejs.com/webpack/building/code-splitting/ )

Split point are created like it should:
screen shot 2017-03-21 at 4 38 36 pm

And you can see them being loaded in the Network Devtools when navigating.

Dynamic import provides less functionality than require.ensure. While it's possible to handle errors with it, features like naming are available for require.ensure only.

Great boilerplate guys 👍


Elyx0:
Since this PR covers https://github.com/chentsulin/electron-react-boilerplate/issues/832#issuecomment-287568603 too I've tried everything but it seems flow doesn't like the dynamic `import()` thus raising error. It's present in babel-eslint since 7.1 ( https://github.com/babel/babel-eslint/pull/413 )
I've checked: https://github.com/benmosher/eslint-plugin-import/issues/700 and it seems it comes from flow: https://github.com/facebook/flow/issues/2968

Should we ignore the lines with these declarations?

amilajack:
> Should we ignore the lines with these declarations?

That would be a good idea for the time being. Also do you know why the end to end testes are failing?

Elyx0:
Got it, I think it has to do with the new router that needs wrapping the provider of the test CounterPage because is contains `<Link>` https://github.com/ReactTraining/react-router/issues/1660

.babelrc
```diff
@@ -4,7 +4,7 @@
     "stage-0",
     "react"
   ],
-  "plugins": ["add-module-exports"],
+  "plugins": ["add-module-exports","dynamic-import-webpack"],
```
amilajack: Is this necessary? Webpack 2 natively supports dynamic imports

app/containers/Bundle.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import { Component } from 'react';
+import type { Children } from 'react';
+
+type Props = {children: Children, load: () => Promise<*>};
+class Bundle extends Component {
```
amilajack: nit: convention is to use a new line before class declarations

app/containers/Bundle.js
```diff
@@ -0,0 +1,45 @@
+// @flow
```
amilajack: What exactly is this `Bundle` container necessary? Some inline comments explaining why this is necessary would be nice

app/containers/Bundle.js
```diff
@@ -0,0 +1,45 @@
...
+    this.load(this.props);
+  }
+
+  componentWillReceiveProps(nextProps: Props) {
+    if (nextProps.load !== this.props.load) {
+      this.load(nextProps);
+    }
+  }
+
+  props: Props
```
amilajack: nit: Type declarations should come before methods and properties

package.json
```diff
@@ -119,6 +119,7 @@
     "babel-loader": "^6.2.10",
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-dev-expression": "^0.2.1",
+    "babel-plugin-dynamic-import-webpack": "^1.0.1",
```
amilajack: Not sure if this is necessary

amilajack:
Closing in favor of https://github.com/chentsulin/electron-react-boilerplate/pull/884

PR #843, opened 2017-03-22
Adding plop support for ease of dev

Unforgiven-wanda:
Hello,
I've been using this boilerplate for some time now, and something that's always bothered me was the lack of CLI actions to quickly create components, especially so for stateful components as you have to create multiple files and edit some others. It was tedious.

I stumbled yesterday across `plop` and lo and behold! With this modification, you can `npm run plop` to create a new component or helper. The files are created automatically and the `reducers/index.js` is edited as well, getting you started immediately.

You can either create a helper file in the utils folder, a styled or not stateless component, or a styled or not stateful component.

All that's left is to edit the `configureStore.development.js` if a stateful component is added, but that's all I have time for today, and I'm not even sure this pull request will be accepted, so... :)

What's more, the `plopfile` can be extended indefinitely by the contributors or the users themselves.
Hope this helps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "package-win": "npm run build && build --win --x64",
     "package-linux": "npm run build && build --linux",
     "package-all": "npm run build && build -mwl",
-    "cleanup": "mop -v"
+    "cleanup": "mop -v",
+    "plop": "node ./node_modules/plop/src/plop.js"
   },
   "browserslist": "electron 1.4",
   "build": {
--- a/plopfile.js
+++ b/plopfile.js
@@ -1,9 +1,22 @@
 module.exports = function (plop) {
   // create your generators here
-  plop.setGenerator('basics', {
-    description: 'this is a skeleton plopfile',
-    prompts: [],
-    actions: []
+  plop.setGenerator('helper', {
+    description: 'Create a utils file',
+    prompts: [{
+      type: 'input',
+      name: 'name',
+      message: 'What is the name of your file?',
+      validate: function (value) {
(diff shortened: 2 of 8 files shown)


amilajack:
Would like some links that I can see so I can learn more about plop

Unforgiven-wanda:
@amilajack Here https://github.com/amwmedia/plop
The gif is pretty self-explanatory too.

amilajack:
@Unforgiven-wanda I'm somewhat reluctant to add dependencies on third party code libs I haven't used before. Is there another boilerplate or a large scale project that uses this?

Unforgiven-wanda:
@amilajack No, none that I know of anyway.

amilajack:
The issue with this is what if one of the maintainers updates a file and forgets to change the plop file template?

Unforgiven-wanda:
Basically now what plop does is to generate files, for this use case a new stateful component. And given that for it to be useful you need to edit the reducer's index I figured that yes, it should do it automatically. What we could have however, in the case you mention (the user edits this file manually, causing the regex to fail) is a more precise regex (To make sure the file hasn't been edited manually), that, should it fail, would let the user know.
Something like:
-> Add component
-> Edit `reducers/index.js` file to add our new component, on fail display message "Warning: Automatic edit failed. You should edit the `reducers/index.js` yourself.
...Something along those lines.

And even if we don't, you still would have a nifty CLI tool to generate files easily.

amilajack:
Closing this as a more elegant solution presents itself. I'll be on the lookout for a tool like this that has traction

PR #860, opened 2017-03-29
Initial Webpack DLL Support

amilajack:
- [x] Enables Webpack DLL for faster Hot Reloading/Reloading


.babelrc.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * @HACK: Babel doesn't support using modules as your `babelrc`.
```
jhen0409: Maybe just edit `.babelrc` instead of new file? [Like my usage](https://github.com/jhen0409/react-native-debugger/blob/master/webpack/base.babel.js).

Also, we can waiting babel 7 for [`.babelrc.js` support](https://github.com/babel/babel/pull/4892).

amilajack: The issue with this is that node-register doesn't support imports. So none of the npm scripts will work because the import statements won't be transpiled. Use your solution of reading and parsing the `.babelrc`

amilajack:
@jhen0409 couldn't get `modules: false` to work without breaking hot reloading. Lets save that for another PR

PR #866, opened 2017-03-30
More flexible NODE_ENV with DefinePlugin

amilajack:


johnryan:
@amilajack Should these changes also be made in the other webpack configuration files?

PR #880, opened 2017-03-31
Add SASS support to project

Miziak:


amilajack:
I love how this PR adds support for sass out of the box. But I dont want to make knowing sass a barrier to using the boilerplate. How about we keep keep the `.css` files as they were but keep all the other changes this PR brings?

Miziak:
Ok so i revert to original `.css` files and keep my SASS configs as option to use.

amilajack:
Yea. also I would recommend commenting why the sass webpack configs are there. Someone may think they are unnecessary in the future and try to remove them. Comment something like 'these are hear for out of the box support for sass'

amilajack:
Also I forgot to ask. Does this work with live reloading? What's the performance like?

Miziak:
Ok, I reverted all css files to state from master branch and short feature description in README.md.
I also added some feature comments in webpack config files.

amilajack:
Does this work with hot reloading?

Miziak:
Yes it is, reloads view almost instantly so I think it performs well.

amilajack:
Great! Also is it possible to add the webpack configs to the webpack base config? What was the reason for adding to both dev and prod configs?

PR #887, opened 2017-04-01
React Router 4 Migration

amilajack:
Closes https://github.com/chentsulin/electron-react-boilerplate/issues/808 https://github.com/chentsulin/electron-react-boilerplate/issues/833

Remember to mark off https://github.com/chentsulin/electron-react-boilerplate/issues/504


app/routes.js
```diff
@@ -1,14 +1,13 @@
...
+import { Route } from 'react-router';
 import App from './containers/App';
 import HomePage from './containers/HomePage';
 import CounterPage from './containers/CounterPage';
 
-
 export default (
   <Route path="/" component={App}>
-    <IndexRoute component={HomePage} />
+    <Route path="/" component={HomePage} />
```
thisconnect: hi, thanks for the PR! I am playing around with it, but this did give me to following warning (also Homepage didn't show)
```
Warning: You should not use <Route component> and <Route children> in the same route; <Route children> will be ignored
```

I could get around it by removing the path compoent on the root route and using <App> directly. As seen in the first code example here https://teamtreehouse.com/community/warning-you-should-not-use-route-component-and-route-children-in-the-same-route-route-children-will-be-ignored

amilajack: Would be great if you could PR to this branch (react-router-4)

thisconnect: @amilajack I opened a PR on the react-router-4 branch, see https://github.com/chentsulin/electron-react-boilerplate/pull/901

PR #901, opened 2017-04-07
Fix npm start and npm run dev

thisconnect:
Followup https://github.com/chentsulin/electron-react-boilerplate/pull/887#discussion_r110197578

### DIFF
--- a/app/containers/App.js
+++ b/app/containers/App.js
@@ -1,10 +1,10 @@
 // @flow
 import React, { Component } from 'react';
-import type { Element } from 'react';
+import type { Children } from 'react';
 
 export default class App extends Component {
   props: {
-    children: Element<*>
+    children: Children
   };
 
   render() {
--- a/app/containers/App.js
+++ b/app/containers/App.js
@@ -1,10 +1,10 @@
 // @flow
 import React, { Component } from 'react';
-import type { Children } from 'react';
+import type { Element } from 'react';
 
 export default class App extends Component {
   props: {
-    children: Children
+    children: Element<*>
   };
 
   render() {
--- a/app/containers/App.js
+++ b/app/containers/App.js
@@ -1,10 +1,9 @@
 // @flow
 import React, { Component } from 'react';
-import type { Children } from 'react';
 
 export default class App extends Component {
   props: {
-    children: Children
+    children: any
   };
 
   render() {
--- a/app/containers/App.js
+++ b/app/containers/App.js
@@ -1,9 +1,10 @@
 // @flow
 import React, { Component } from 'react';
+import type { Children } from 'react';
 
 export default class App extends Component {
   props: {
-    children: any
+    children: Children // eslint-disable-line
   };
 
   render() {
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -2,7 +2,7 @@
 import { app, BrowserWindow } from 'electron';
 import MenuBuilder from './menu';
 
-/* eslint global-require: 2, flowtype-errors/show-errors: 0 */
+/* eslint global-require: 1, flowtype-errors/show-errors: 0 */
 
 let mainWindow = null;
 
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -2,7 +2,7 @@
 import { app, BrowserWindow } from 'electron';
 import MenuBuilder from './menu';
 
-/* eslint global-require: 1, flowtype-errors/show-errors: 0 */
+/* eslint global-require: 2, flowtype-errors/show-errors: 0 */
 
 let mainWindow = null;
 
--- a/app/routes.js
+++ b/app/routes.js
@@ -1,13 +1,18 @@
 // @flow
 import React from 'react';
-import { Route } from 'react-router';
+import { BrowserRouter as Router } from 'react-router-dom';
+import { Switch, Route } from 'react-router';
 import App from './containers/App';
 import HomePage from './containers/HomePage';
 import CounterPage from './containers/CounterPage';
 
-export default (
-  <Route path="/" component={App}>
-    <Route path="/" component={HomePage} />
-    <Route path="/counter" component={CounterPage} />
-  </Route>
+export default () => (
+  <Router>
+    <App>
+      <Switch>
+        <Route path="/counter" component={CounterPage} />
+        <Route path="/" component={HomePage} />
+      </Switch>
+    </App>
+  </Router>
 );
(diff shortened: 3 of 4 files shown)


app/routes.js
```diff
@@ -1,13 +1,18 @@
...
 import App from './containers/App';
 import HomePage from './containers/HomePage';
 import CounterPage from './containers/CounterPage';
 
-export default (
-  <Route path="/" component={App}>
-    <Route path="/" component={HomePage} />
-    <Route path="/counter" component={CounterPage} />
-  </Route>
+export default () => (
```
amilajack: Can you please elaborate on the changes made here

app/routes.js
```diff
@@ -1,13 +1,18 @@
 // @flow
 import React from 'react';
-import { Route } from 'react-router';
+import { BrowserRouter as Router } from 'react-router-dom';
```
amilajack: What does this change do?

thisconnect:
probably wont pass due to npm run lint

app/main.development.js
```diff
@@ -2,7 +2,7 @@
 import { app, BrowserWindow } from 'electron';
 import MenuBuilder from './menu';
 
-/* eslint global-require: 1, flowtype-errors/show-errors: 0 */
+/* eslint global-require: 2, flowtype-errors/show-errors: 0 */
```
amilajack: `global-require: 1` is necessary because global `require()` is necessary.

app/routes.js
thisconnect: Sorry I have to run, I'll try to reply tonight or tomorrow

app/routes.js
thisconnect: Related docs https://reacttraining.com/react-router/web/api/Router, we can use BrowserRouter or HashRouter from 'react-router-dom'. Docs about BrowserRouter https://reacttraining.com/react-router/web/api/BrowserRouter

The `{ BrowserRouter as Router }` is simply importing and renaming it to Router, that we can use `<Router>` later on, instead of `<BrowserRouter>`. This is optional we could use `<BrowserRouter>` if you prefer that.

app/routes.js
thisconnect: This now exports a function that returns the Router construct, not sure why it errors if the React.createElement is directly exported (see error below). When loggin the function it will only log once unless something is changed and HMR kicks in. In that case the router function is called again. HMR seems to work as far as I tested.

screen shot 2017-04-08 at 9 50 28 am

app/routes.js
```diff
@@ -1,13 +1,18 @@
...
 import CounterPage from './containers/CounterPage';
 
-export default (
-  <Route path="/" component={App}>
-    <Route path="/" component={HomePage} />
-    <Route path="/counter" component={CounterPage} />
-  </Route>
+export default () => (
+  <Router>
+    <App>
```
thisconnect: I was pretty sure that `<Router component={App}>` should now be `<Router><App>` as it is always within App, but both styles seem to work. Do prefer `<Router component={App}>`  ?

app/routes.js
```diff
@@ -1,13 +1,18 @@
...
+      <Switch>
+        <Route path="/counter" component={CounterPage} />
+        <Route path="/" component={HomePage} />
```
thisconnect: I'm not happy how the path are defined. I expected to be able to use `<Route exact path="/" component={Homepage}>` but location.pathname is '/Users/path/electron-react-boilerplate/app/app.html'. This would work but feels wrong: `<Route path="*/app.html" component={HomePage} />`

package.json
```diff
@@ -17,6 +17,7 @@
     "build-main": "cross-env NODE_ENV=production node --trace-warnings -r babel-register ./node_modules/webpack/bin/webpack --config webpack.config.main.prod.js --progress --profile --colors",
     "build-renderer": "cross-env NODE_ENV=production node --trace-warnings -r babel-register ./node_modules/webpack/bin/webpack --config webpack.config.renderer.prod.js --progress --profile --colors",
     "start": "cross-env NODE_ENV=production electron ./app/",
+    "prestart": "npm run build",
```
thisconnect: I can remove this, as it is documented in the readme. On the other hand I would expect `npm install` and `npm start` to "just work". But the readme clearly says you have to do `npm run build` before. Should I undo the change here?

thisconnect:
@amilajack I added some comments, as you can see I do not have 3 years of react-router4 experience I was just playing around. Nevertheless I hope it helps.

thisconnect:
Now the linter complains about React element App:  property `children`.

https://travis-ci.org/chentsulin/electron-react-boilerplate/jobs/219967223#L3139

I honestly have no idea about flow or is it flow-typed? Could not find any documentation about it :/

package.json
amilajack: Lets keep it. Good idea on how the `npm start` should "just work"

app/routes.js
amilajack: Do you know if the children props are necessary anymore?

thisconnect: The line in question was https://github.com/thisconnect/electron-react-boilerplate/blob/4950ad98afcf25a2ae7734d5406c76abc345ea8e/app/containers/App.js#L7

I'm not exactly sure where to look for react flow types, especially children. My personal opinion is that tools like flow should help the developer. However I find it quite hard to google this particular lint warning and am not sure where to look. flow, flow-type something lint?
No mention about children https://flow.org/en/docs/frameworks/react/

This issue came up in my search but didn't help https://github.com/facebook/flow/issues/1355

This seems to be the type definition for react https://github.com/facebook/flow/blob/master/lib/react.js#L157

My last commit does lint now and it looks right to me https://github.com/thisconnect/electron-react-boilerplate/commit/ced774dc86241004ae0e80a8aac8f2e85ee1387c

package.json
thisconnect: 👍

thisconnect:
Not sure why it still fails linting (see https://travis-ci.org/chentsulin/electron-react-boilerplate/jobs/220217180#L2965). Locally `npm run lint` does pass for me. I just did a `npm install` to make sure that I have got the latest deps. BTW: quite a few deps are not up to date `npm outdated`

thisconnect:
https://github.com/chentsulin/electron-react-boilerplate/blob/react-router-4/.travis.yml#L52-L56 runs fine for me locally

app/routes.js
amilajack: [This repo](https://github.com/thejameskyle/codemod-react-proptypes-to-flow) might be useful. For now, try using `eslint-disable-line` to disable that line from being reported and Above that line, write a comment saying the issue you were having.

app/containers/App.js
```diff
@@ -1,9 +1,10 @@
 // @flow
 import React, { Component } from 'react';
+import type { Children } from 'react';
 
 export default class App extends Component {
   props: {
-    children: any
+    children: Children // eslint-disable-line
```
amilajack: I think the comment belongs on `app/routes.js ` on line `11`

PR #907, opened 2017-04-11
Deleted reference to erb-logo.png of setup.js

gloob:
Deleted reference to erb-logo.png file in remove array of boiler-room-custodian setup file because it is not present in the project.


amilajack:
Thanks!

PR #916, opened 2017-04-15
Made HtmlWebpackPlugin inject bundles to app.html

williamoliveira:


williamoliveira:
Could app.html being served from webpack-dev-server instead of filesystem break things?

amilajack:
I'll look into this. I don't think it will break this boilerplate but it will likely be a breaking change for larger apps that may have hardcoded routes

PR #917, opened 2017-04-15
flow-typed install to postinstall + add any exports for module(require)

VladimirPal:
As a new user i tried this boilerplate but found few flow issues.


amilajack:
See https://flow.org/blog/2016/10/13/Flow-Typed/

> This is the reason we advise that you commit your installed libdefs to version control rather than rely on a system like npm+semver to download and install a non-deterministic semver-ranged version from npm. Checking in your libdefs ensures that all collaborators on your project have consistent output from Flow at any given commit in version history.

VladimirPal:
@amilajack In this case you need to freeze packages versions, because if you will try to install all dependencies from scratch you willl see flow errors.

amilajack:
Got it. So updating the flow types will fix this? I think a better solution would be to update the types in more frequent intervals. Thoughts on this?

VladimirPal:
@amilajack Updating the flow will fix errors with redux lib, but stylelint and babel-plugin-flow-runtime still yield errors, so i added it to ignore list, also flow can't find module.js(And how!), i've added some kind of hack https://github.com/web-pal/electron-react-boilerplate/blob/master/flow-typed/npm/module_vx.x.x.js

Anyway if you don't want new users will have flow errors we need to freeze dependencies versions, or at least mention in documentation about this issue(something like, if you see flow errors in node_modules, try to reinstall flow-typed).

But i think if you don't have freezed dependencies, this advice makes no sense and we can use postinstall.
> This is the reason we advise that you commit your installed libdefs to version control rather than rely on a system like npm+semver to download and install a non-deterministic semver-ranged version from npm. Checking in your libdefs ensures that all collaborators on your project have consistent output from Flow at any given commit in version history.

amilajack:
Thanks!

Issue #909, opened 2017-04-12
Live reload loads old content


williamoliveira:
When it does full page reload it loads with the old version
I dont know how to explain it better so I made a quick video to demonstrate: https://youtu.be/xk-4Q0YAIlw

amilajack:
Interesting. Also thanks for the recording. Makes it so much easier for me to see what's going on on your end. If you haven't change anything in the boilerplate after cloning it, it may be because your dependencies are outdated or mismatching. Using yarn will most likely fix this kind of issue:

```
git pull --all
npm i -g yarn
rm -rf node_modules
yarn
```

williamoliveira:
Did not fix
It is a clean clone and I was already using yarn

amilajack:
try cleaning the yarn cache
```
yarn cache clean
```

and then repeat the previous steps

aemxdp:
Same issue for me. Started using `electron-react-boilerplate` few hours ago.

williamoliveira:
Cleaned yarn cache, the problem persists

Slapbox:
Are you using any node native modules? An example would be `electron-spellchecker`

This module may have been the cause of some issues I've had, but I cannot confirm this is the cause yet.

anthonyraymond:
I'm also afected since i rebased from #856 to #898. It may be related to DLL config, since i also often encounter #905 on reload

Slapbox:
@anthonyraymond See my comments on #905. I will continue to update based on my findings.

amilajack:
Would be great if everyone here could test out https://github.com/chentsulin/electron-react-boilerplate/pull/914

williamoliveira:
No luck
But it seems to fix #905 that I was having too sometimes

amilajack:
@williamoliveira thats the proper solution for this. Would be great if you could PR to use HtmlWebpackPlugin instead. I don't have much time on my hands this week

williamoliveira:
I will try

williamoliveira:
I managed to get HtmlWebpackPlugin working, I will do some more manual testing and PR it after
But unfortunately it did not fix this issue :/

amilajack:
Can someone please confirm that https://github.com/chentsulin/electron-react-boilerplate/pull/916 resolves the race condition issue? I am not experiencing the race condition issue myself so its impossible to me to test this

williamoliveira:
It did not, but fixes #905 at least

[...]

PR #920, opened 2017-04-17
Fixed HMR (disabled multiStep)

uppe-r:
https://github.com/jantimon/html-webpack-plugin/issues/533 Fixes Live reloading loading old content (#909)


amilajack:
Made some changes. Would be great if you could test again and merge

PR #945, opened 2017-04-23
Use local flow-bin package

talha131:
If this setting is not present then flow extension suggests to install
flow globally which should is not necessary at all because locally
flow-bin is already available.


amilajack:
Thanks!

Issue #935, opened 2017-04-21
In Packaged Application: Cannot read property 'dispatch' of undefined


Slapbox:
```
Uncaught (in promise) TypeError: Cannot read property 'dispatch' of undefined
    at ConnectedRouter._this.handleLocationChange (ConnectedRouter.js:26)
    at ConnectedRouter.componentWillMount (ConnectedRouter.js:41)
    at ReactCompositeComponentWrapper.performInitialMount (ReactCompositeComponent.js:351)
    at ReactCompositeComponentWrapper.mountComponent (ReactCompositeComponent.js:258)
    at Object.mountComponent (ReactReconciler.js:46)
    at ReactCompositeComponentWrapper.performInitialMount (ReactCompositeComponent.js:371)
    at ReactCompositeComponentWrapper.mountComponent (ReactCompositeComponent.js:258)
    at Object.mountComponent (ReactReconciler.js:46)
    at ReactCompositeComponentWrapper.performInitialMount (ReactCompositeComponent.js:371)
    at ReactCompositeComponentWrapper.mountComponent (ReactCompositeComponent.js:258)

```
```

ConnectedRouter._this.handleLocationChange	@	ConnectedRouter.js:26
componentWillMount	@	ConnectedRouter.js:41
performInitialMount	@	ReactCompositeComponent.js:351
mountComponent	@	ReactCompositeComponent.js:258
mountComponent	@	ReactReconciler.js:46
performInitialMount	@	ReactCompositeComponent.js:371
mountComponent	@	ReactCompositeComponent.js:258
mountComponent	@	ReactReconciler.js:46
performInitialMount	@	ReactCompositeComponent.js:371
mountComponent	@	ReactCompositeComponent.js:258
mountComponent	@	ReactReconciler.js:46
performInitialMount	@	ReactCompositeComponent.js:371
mountComponent	@	ReactCompositeComponent.js:258
mountComponent	@	ReactReconciler.js:46
performInitialMount	@	ReactCompositeComponent.js:371
mountComponent	@	ReactCompositeComponent.js:258
mountComponent	@	ReactReconciler.js:46
mountComponentIntoNode	@	ReactMount.js:104
perform	@	Transaction.js:140
batchedMountComponentIntoNode	@	ReactMount.js:126
perform	@	Transaction.js:140
batchedUpdates	@	ReactDefaultBatchingStrategy.js:62
batchedUpdates	@	ReactUpdates.js:97
[...]

PR #949, opened 2017-04-24
Fix router

anthonyraymond:
Since `ConnectedRouter` has been introduced, `history.push` `history.goBack` and any other`history` function has no effect on the rendering process (but dispatch actions properly ).

This probably happens because there are two router nested `HashRouter` from *app/routes.js* is nested into `ConnectedRouter` from *app/containers/Root.js*

How to reproduce:
**app/Home.js**
```
// @flow
import React, { Component } from 'react';
import styles from './Home.css';
import { history } from '../store/configureStore';

export default class Home extends Component {
  render() {
    return (
      <div>
        <div className={styles.container} data-tid="container">
          <h2>Home</h2>

          <button onClick={() => { history.push('/counter'); }}>
            test
          </button>
        </div>
      </div>
    );
  }
}

```


amilajack:
@anthonyraymond changed the base of the PR to release `v0.11.2`. Would be great if you could resolve the merge conflicts

anthonyraymond:
Should be ok. I used the online user-friendly "Resolve button". Didn't knew that ever exists x)
I hope this will work. Otherwise i'll do a manual commit tonight

amilajack:
@Slapbox Would really appreciate if you could test this out before merging into `v0.11.2`

Slapbox:
I implemented it without issue in my app, but I'm not actively using the router. I am 100% single paged, so my test case is very weak, just loading my main page and then not routing again.

Slapbox:
By the way, since it's related to router, worth mentioning it has no effect on #935 for me.

amilajack:
This is the most annoying thing about the JS ecosystem (and probably other language ecosystems). You change on upgrade some dependency and you find out later that something broke (and you have no idea why).

Slapbox:
Much truth. Dependency hell has been made a lot less terrible by Yarn, but it still exists in its own way.

amilajack:
Sorry for the delay on this. Wanted to take the time to do some testing myself

anthonyraymond:
No problem :)
Thabks for the merge

Issue #927, opened 2017-04-19
PR: Add prettier support/integration


amilajack:

amilajack:
Closing this because prettier and eslint integration and consistency is off

loopmode:
@amilajack Actually, today it seems to be working together nicely once ESLint is configured accordingly.

This is all to make it work: https://github.com/loopmode/electron-react-boilerplate/commit/72bfdec6d1f4e360849cc6df4c3382da29638c7c

This is the result after `npm run lint-fix`: https://github.com/loopmode/electron-react-boilerplate/commit/78e38c02f43faecec35454090dee2d32d038a8fb (80 chars/row)

However, I guess when using prettier, one should clean up the ESLint setup, e.g. drop `eslint-formatter-pretty`. I've no idea whether currently, double work is done as in maybe first `eslint-formatter-pretty` does its thing, then comes `prettier` and does its thing, making the first step obsolete...?

Issue #919, opened 2017-04-16
PR: Folder structure refactor


ansumanshah:
Proposal for changes in folder structure (if possible)
1. Can we move all the webpack config files to one webpack folder?
2. Move all the flow, mocks, dll, and webpack folders under config
3. Move resources and img under assets

Would be happy to submit a PR

amilajack:
the `flow-typed` directory was removed recently. I think the folder should be called `webpack` to be more explicit. Other than that, 👌 Would love a PR for this

ansumanshah:
so I was suggesting a structure same as react-boilerplate with the `internals` folders they have, is that what you have in mind as well?

ansumanshah:
Can we do away with the `img` folder as well please, It was confusing from me when I started

ansumanshah:
I am also proposing to do away with the config files we have for stylelint, eslint and babel to either move them to `package.json`, webpack or create config files/folder for them in the internals folder.

P.S would love to work on the site we are planning as well, can we move both the repos to a separate org?

amilajack:
> Can we do away with the img folder as well please, It was confusing from me when I started

Yup! I was actually thinking the same

PR #952, opened 2017-04-25
Folder Refactor

ansumanshah:
#919 #927 

Would like to make this repo more beginner friendly, so would suggest merging some more folders to one and removing some of the redundant config files.

- [x] Move the webpack, flow, mocks folder to config/internals (just like `react-boilerplate`)


amilajack:
Let's try adding the prettier integration in another PR. Reviewing this is difficult because there are so many changes. Also see [my comment ](https://github.com/chentsulin/electron-react-boilerplate/issues/919#issuecomment-297114525)

config/webpack/webpack.config.renderer.dev.dll.js
```diff
@@ -6,8 +8,9 @@ import webpack from 'webpack';
 import path from 'path';
 import merge from 'webpack-merge';
 import baseConfig from './webpack.config.base';
-import { dependencies } from './package.json';
 
+const pkg = require(path.resolve(process.cwd(), 'package.json'));
```
amilajack: What is the reason for removing the `dependencies` import?

config/webpack/webpack.config.main.prod.js
```diff
@@ -17,7 +17,7 @@ export default merge.smart(baseConfig, {
 
   // 'main.js' in root
   output: {
-    path: __dirname,
```
amilajack: What are the benefits of using `process.cwd()` instead of `__dirname`?

config/webpack/webpack.config.base.js
```diff
@@ -1,10 +1,14 @@
...
+
 /**
  * Base webpack config used across other specific configs
  */
 
 import path from 'path';
 import webpack from 'webpack';
-import { dependencies as externals } from './app/package.json';
+
+const pkg = require(path.resolve(process.cwd(), './app/package.json'));
```
amilajack: What is the reason for removing the `dependencies` import?

config/webpack/webpack.config.main.prod.js
ansumanshah: `__dirname` will return the directory containing this file and not the root folder, so by using `process.cwd()` we make sure that the paths are relative to the folder where the node command is run from.

config/webpack/webpack.config.renderer.dev.dll.js
ansumanshah: I have not removed the dependencies import its just instead of using `../../package.json`, I preferred using the path from the current working directory and relating `package.json` to it, so that wherever this file maybe we will not have to change the import statement again and again.

config/webpack/webpack.config.base.js
ansumanshah: Same as the last comment, I think this should also be linked to the root directory and not use the relative paths like `../../app/package.json`

ansumanshah:
am I missing something here?

amilajack:
I actually deleted the dev-master branch and I think GitHub closed this on my behalf.

amilajack:
Changed the base of the PR to `v0.11.2`

config/webpack/webpack.config.base.js
amilajack: @ansumanshah I think the `import` method is much more idiomatic. For now let's keep the paths as they were. We might change this later.

ansumanshah:
Done with the changes you recommended
@amilajack what are these commits and conflicts recently added, and what's the best way to resolve this?

amilajack:
@ansumanshah I would recommend creating a new PR and applying only the changes that are necessary.

ansumanshah:
I believe this is just with webpack folders refactor only, it's just that the base branch is not synced with the current one.

amilajack:
@ansumanshah yea. Sorry for the confusion about which branch to merge into. We now use the most recently release branch as the branch we merge into. Right now, that's `v0.12.0`

amilajack:
Went ahead and did this in https://github.com/chentsulin/electron-react-boilerplate/pull/970

PR #997, opened 2017-05-17
Make runTests.js exit non-zero when tests fail

midgethoen:
runTests.js always exits status zero, causing CI not to notice when tests fail.


amilajack:
@midgethoen thanks for pointing this out! How do you recommend i test this out locally?

midgethoen:
I recommend adding a failing test, running the tests and echo the exit code.
```
npm run test-e2e
echo $?
```

amilajack:
Thanks!

PR #1008, opened 2017-05-23
fix react router 4 should not use <Route component> and <Route childr…

aiwb:
fix warning: react router 4 should not use ```<Route component>``` and ```<Route children>``` in the same route; ```<Route children>``` will be ignore.


amilajack:
@aiwb Thanks! Does this happen to fix https://github.com/chentsulin/electron-react-boilerplate/issues/1007?

PR #1023, opened 2017-05-27
v0.12.0

amilajack:
- [x] Removed `babel-polyfill`
- [x] Renamed and alphabetized npm scripts
- [x] Changed node dev `__dirname` and `__filename` to node built in fn's (https://github.com/chentsulin/electron-react-boilerplate/pull/1035)
- [x] Enable node_modules cache on CI
- [x] Rename `app/bundle.js` => `app/renderer.prod.js`, `dll/vendor.js` => `dll/renderer.dev.dll.js`

Closes https://github.com/chentsulin/electron-react-boilerplate/issues/1043
Closes https://github.com/chentsulin/electron-react-boilerplate/issues/1012
Closes https://github.com/chentsulin/electron-react-boilerplate/issues/1109


webpack.config.renderer.dev.js
```diff
@@ -50,6 +50,23 @@ export default merge.smart(baseConfig, {
   module: {
     rules: [
       {
+        test: /\.jsx?$/,
+        exclude: /node_modules/,
+        use: {
+          loader: 'babel-loader',
+          options: {
+            cacheDirectory: true,
+            plugins: [
```
jhen0409: Maybe `class-properties` and `es2015-classes` plugins can also included here then we can add comments for why they are used?

amilajack: We still need those plugins for the main process

jhen0409: Should not be right, `stage-0` included `class-properties` and main process (node 7.x) doesn't need `es2015-classes`, it just a fix for ensure it put before of `react-hot-loader/babel`.

amilajack: Got it. So we should move `es2015-classes` to the renderer webpack config and remove `class-properties`?

jhen0409: We should keep the `class-properties` before `es2015-classes` otherwise it will throw `Missing class properties transform` error. (plugins run before presets, we can see in [babel doc](https://babeljs.io/docs/plugins/#plugin-preset-ordering))

amilajack: @jhen0409 just pushed the new changes. 👍 or 👎 ?

jhen0409:
I'm experiencing the error after removed `es2015-classes` in main process:

```
SyntaxError: /Users/jhen/workspace/electron-react-boilerplate/app/menu.js:
  Decorators are not officially supported yet in 6.x pending a proposal update.
```

I think it may a bug for babel 6, it thrown even if we haven't use decorators. It looks like babel 7 (alpha) have been added back the [`decorators`](https://github.com/babel/babel/tree/7.0/packages/babel-plugin-transform-decorators) plugin, so I think currently the workaround is add `transform-decorators-legacy` to babelrc.

elsueno:
I can't find the commit that closes #1109 here. As I currently can't update to v0.12.0 because I can't get transform-decorators-legacy to work I just wanted to cherry-pick the fix for #1109.
Thanks!

PR #1027, opened 2017-05-30
Fix npm command in README.md

jameskraus:
Fixed the npm command which was spelled incorrectly.


jameskraus:
NVM, that's actually a thing that exists

amilajack:
`apm` is short for atom package manager. This is not a mistake

PR #1031, opened 2017-05-31
Fixed npm start script importing ES6 modules

### LINKED ISSUE #996: npm start does not run with babel-register
npm start is inconsistent by omitting -r babel-register electron parameter

package.json should be fixed to:
    "start": "cross-env NODE_ENV=production electron -r babel-register ./app/"

amilajack:
Closes #996


jhen0409:
We shouldn't enable babel transform in production, it still break `remote.require` usage in package even if we used this patch (meaning [this comment](https://github.com/chentsulin/electron-react-boilerplate/issues/996#issuecomment-303638411)), although I'm not sure if it really solved the `remote.require` problem.

I think it should be a limitation for use webpack (I usually avoid using it), we might can report this problem to [webpack](https://github.com/webpack/webpack) for solving this problem.

PR #1048, opened 2017-06-09
Add more details to the Module Structure

talha131:
See discussion here
https://github.com/chentsulin/electron-react-boilerplate/issues/1042#issuecomment-307514773

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,13 +191,33 @@ $ npm run test-e2e
 
 See [electron-builder CLI Usage](https://github.com/electron-userland/electron-builder#cli-usage)
 
-#### Module Structure
+## How to add modules to the project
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,14 +191,34 @@ $ npm run test-e2e
 
 See [electron-builder CLI Usage](https://github.com/electron-userland/electron-builder#cli-usage)
 
-#### Module Structure
+## How to add modules to the project
```
amilajack: How about `## How to install dependencies in the boilerplate`?

talha131: JS ecosystem is not consistent on the term. They use module/dependency/package interchangeably. For example,

> yarn add <package...>
> This will install one or more packages in your dependencies.

> yarn install
> Install all the dependencies listed within package.json in the local node_modules folder.

node-gyp documentation uses addon.

> node-gyp is a cross-platform command-line tool written in Node.js for compiling native addon modules for Node.js.

Finally, we add edit "dependency" section of "package".json, which finally installs in "modules" directory :)

**You suggest the term and I will update wiki and readme to be consistent.**

**I vote for `package`.** I think it is more common in docs than the terms modules and dependency. For example,

> Material-UI is available as an npm package.

PR #1052, opened 2017-06-12
Removed babel-polyfill

amilajack:
`babel-polyfill` is no longer needed. We're using `babel-preset-env`'s `useBuiltIns` option to load only the polyfills that are required by the specific node version. This PR trims the build size rather significantly.


amilajack:
@jhen0409 would love your feedback on this

package.json
```diff
@@ -201,8 +200,8 @@
     "source-map-support": "^0.4.15"
   },
   "devEngines": {
-    "node": ">=6.x",
-    "npm": ">=3.x",
+    "node": ">=7.x",
+    "npm": ">=5.x",
```
jhen0409: Should we use `>=4.x` as node 7.x builtin?

amilajack: Good call!

Issue #1065, opened 2017-06-15
"npm run package" not working


jfcartier:
Ok so I followed the doc, step by step. 

$ git clone --depth=1 https://github.com/chentsulin/electron-react-boilerplate.git your-project-name
$ cd your-project-name
$ npm install
$ npm run dev 

So far everythings fine. The errors come when I run the package script.

$ npm run package

Here's my non exhaustive logs below. Do you have the same issue?

Thanks 

```
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! electron-react-boilerplate@0.12.0 build: `concurrently "npm run build-main" "npm run build-renderer"`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the electron-react-boilerplate@0.12.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/jfcartier/.npm/_logs/2017-06-15T14_14_08_946Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! electron-react-boilerplate@0.12.0 package: `npm run build && build --publish never`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the electron-react-boilerplate@0.12.0 package script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/jfcartier/.npm/_logs/2017-06-15T14_14_08_965Z-debug.log
```
File logs : https://cl.ly/2U0M1a11263D

amilajack:
Try running `npm run build-main` and `npm run build-renderer` separately. Do any of them run successfully?

jfcartier:
Hi @amilajack, so it actually crash when I launch "npm run build-main". 
I think the main error is this 

**ERROR in unknown: Don't use `path.replaceWith()` with a source string, use `path.replaceWithSourceString()`**

I've seen this error on "npm install" too I think. 
Here's full log below. 

Thanks

```
~/Desktop/allo $ npm run build-main

> electron-react-boilerplate@0.12.0 build-main /Users/jfcartier/Desktop/allo
> cross-env NODE_ENV=production node --trace-warnings -r babel-register ./node_modules/webpack/bin/webpack --config webpack.config.main.prod.js --progress --profile --colors

1525ms building modules
60ms sealing
[...]

PR #1072, opened 2017-06-16
Use Yarn instead of npm install

jfcartier:
There's a problem with npm install and the option to package afterward. With yarn, you don't have this. I know it may be another required technology, but we need to change the doc if we don't fix the problem soon.

(see issue #1065)


amilajack:
Thanks!

PR #1086, opened 2017-06-21
Include `test/setup.js` before each test run.

Fer0x:
After migration to jest, file `test/setup.js` is no longer included before tests run, so all mocks are unavailable. This PR fix that.
Also, updated using of JSDOM and written unit test to make sure, that setup.js is included and mock of localStorage is available.

### DIFF
--- a/test/setup/setup.spec.js
+++ b/test/setup/setup.spec.js
@@ -1,5 +1,0 @@
-describe('setup', () => {
-  it('should mock localStorage', () => {
-    expect(typeof localStorage).toBe('object');
-  });
-});
--- a/test/setup/setup.spec.js
+++ b/test/setup/setup.spec.js
@@ -1,0 +1,5 @@
+describe('setup', () => {
+  it('should mock localStorage', () => {
+    expect(typeof localStorage).toBe('object');
+  });
+});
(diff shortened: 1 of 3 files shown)


test/setup/setup.spec.js
```diff
@@ -0,0 +1,5 @@
+describe('setup', () => {
```
amilajack: Can you please explain why this file is necessary. is localStorage necessary for tests?

Fer0x: It's just checks that `test/setup.js` is included and mocking globals is working. So, if any migration to another unit testing library will be in the future, you will never miss to require `test/setup.js` :)

Fer0x: so?

amilajack: I'm not sure if this file is necessary. We're leaning towards a "minimal configuration" with regard to the boilerplate's setup

Fer0x: Ok, I removed it.

Fer0x:
I removed unnecessary file.

test/setup/setup.spec.js
Fer0x: Is now everything is ok?

Fer0x:
bump

Fer0x:
Close because `test/setup.js` now removed in master.

PR #1116, opened 2017-07-03
Redux-persist

Unforgiven-wanda:
Added Redux-persist to persist app state to disk, and Redux-action-buffer, to show a "loading" page while the state is being rehydrated.

PS: don't mind the plop commits, I made the mistake to reapply these old commits.


anthonyraymond:
Sounds a bit specific, don't you think? In none of my desktop apps i want the state to be persisted.

amilajack:
That was what I was thinking. But I think we should add this to some kind of documentation.

PR #1126, opened 2017-07-12
Pass arguments to jest cli

Fer0x:
Pass arguments to jest cli.
With that you can use it for check coverage like `npm test -- --coverage` or update snapshots like `npm test -- --u`.

Closes https://github.com/chentsulin/electron-react-boilerplate/issues/929


amilajack:
Thanks for this!

PR #1190, opened 2017-08-31
Main and misc modules update

proclaim:
warning and errors are getting annoying when this repo is first installed through `yarn`, even if `npm run build` there are warning and errors (and deprecated module).

these update should be able to remove these.


amilajack:
This was done in `v0.13.0`

PR #1246, opened 2017-10-01
Check and warn the user if there are native deps installed in root folder

### LINKED ISSUE #1171: Warn user when installing native dependency to `./package.json`
Throw a warning message explaining that webpack does not work with native dependencies. Provide the link to further documentation. Provide directions on how to properly install the native dependency. This validation should run as a `postinstall` event.

```js
fs.readdirSync('node_modules')
  .filter(x => ['.bin'].indexOf(x) === -1)
```

Taken from http://jlongster.com/Backend-Apps-with-Webpack--Part-I

uppe-r:
Closes #1171 .

Checks if there are any native dependencies not installed inside the `app` folder as they should be ([https://github.com/chentsulin/electron-react-boilerplate/wiki/Module-Structure----Two-package.json-Structure](url)) on `postinstall`. If there are, a warning is displayed mentioning the issue and a link to the before mentioned page.

By the way, looks like `node-sass` should also be installed inside of `app` by default since it's a native dep.

### DIFF
--- a/internals/scripts/CheckNativeDep.js
+++ b/internals/scripts/CheckNativeDep.js
@@ -1,0 +1,21 @@
+// @flow
+import fs from 'fs';
+import chalk from 'chalk';
+
+const EndOfLine = require('os').EOL;
+const { execSync } = require('child_process');
+
+const locateNativeDeps = async () => {
+  let nativeDeps = [];
+  fs.readdirSync('node_modules')
+    .filter(folder => fs.existsSync(`node_modules/${folder}/binding.gyp`))
+    .forEach(dep => {
+      nativeDeps.push(dep);
+    });
+  return nativeDeps;
+};
+
+locateNativeDeps()
+  .then(nativeDeps => {
+    if (nativeDeps.length > 0) console.log(`${chalk.whiteBright.bgRed.bold(`Webpack does not work with native dependencies.`)} ${chalk.inverse(nativeDeps)} should be installed inside of the app folder. ${EndOfLine} ${EndOfLine}Read more about native dependencies at https://github.com/chentsulin/electron-react-boilerplate/wiki/Module-Structure----Two-package.json-Structure ${EndOfLine}`);
+});
--- a/internals/scripts/CheckNativeDep.js
+++ b/internals/scripts/CheckNativeDep.js
@@ -2,16 +2,27 @@
 import fs from 'fs';
 import chalk from 'chalk';
 
-const locateNativeDeps = async () => fs.readdirSync('node_modules')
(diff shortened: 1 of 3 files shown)


amilajack:
node-sass is a native dep but its used only in development. Doesn't make sense to package it in `./app`

internals/scripts/CheckNativeDep.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import fs from 'fs';
+import chalk from 'chalk';
+
+const EndOfLine = require('os').EOL;
+const { execSync } = require('child_process');
```
amilajack: Let's use ES6 imports for all of these imports.

internals/scripts/CheckNativeDep.js
```diff
@@ -0,0 +1,21 @@
...
+
+const locateNativeDeps = async () => {
```
amilajack: How about this?
```js
const locateNativeDeps = async () => {
  return fs.readdirSync('node_modules')
    .filter(folder => fs.existsSync(`node_modules/${folder}/binding.gyp`))
};
```

internals/scripts/CheckNativeDep.js
```diff
@@ -0,0 +1,21 @@
...
+    .filter(folder => fs.existsSync(`node_modules/${folder}/binding.gyp`))
+    .forEach(dep => {
+      nativeDeps.push(dep);
+    });
+  return nativeDeps;
+};
+
+locateNativeDeps()
+  .then(nativeDeps => {
+    if (nativeDeps.length > 0) console.log(`${chalk.whiteBright.bgRed.bold(`Webpack does not work with native dependencies.`)} ${chalk.inverse(nativeDeps)} should be installed inside of the app folder. ${EndOfLine} ${EndOfLine}Read more about native dependencies at https://github.com/chentsulin/electron-react-boilerplate/wiki/Module-Structure----Two-package.json-Structure ${EndOfLine}`);
```
amilajack: This line doesn't conform to our eslint rules. Please format it accordingly.

uppe-r: Got it, looks like Sublime's linter extension doesn't work that well.... back to Atom!

internals/scripts/CheckNativeDep.js
uppe-r: Agreed, but got any idea on how we'd ignore node-sass or remove it later from the array?

uppe-r: Nvm, got it

internals/scripts/CheckNativeDep.js
uppe-r: Fixed

PR #1259, opened 2017-10-06
Correct Multi Platform Build link in Readme

balcsida:
"Moved to Multi Platform Build on electron.build site."


amilajack:
Thanks!

PR #1411, opened 2018-02-02
Remove jsdom dep

BuckyMaler:
[Jest](https://facebook.github.io/jest/docs/en/configuration.html#testenvironment-string) now ships with jsdom.


amilajack:
Thanks!

PR #1418, opened 2018-02-08
add prettier to format js files

zhaozhiming:
* add prettier, lint-stage, husky and eslint-config-prettier
* add git precommit hook, when commit will use prettier to format git add code
* add `format` npm script, use prettier to format project js code
* use prettier to format existed code


amilajack:
Good stuff! 🎉 Will merge this in soon after some testing

amilajack:
Can you update the `CHANGELOG` with the changes you made? Let's push this under `v0.14.0`

zhaozhiming:
I have not idea which changelog is better.   
@amilajack Could you give some suggestions? 

### A
```markdown
# 0.13.3 (2018.2.8)
- Add git precommit hook, when commit will use prettier to format git add code
- Add `format` npm script which use prettier to format project js code
```

### B
```markdown
# 0.13.3 (2018.2.8)
- Add prettier to format js files
```

amilajack:
A is better

zhaozhiming:
I combed the scripts again.   
And I found prettier is [default ignore `node_modules`](https://prettier.io/docs/en/cli.html#with-node-modules) so no need to write `'*/!(node_modules)/**/*.js'` in script.   
Please review continue. Thanks.

package.json
```diff
@@ -12,6 +12,8 @@
     "electron-rebuild": "electron-rebuild --parallel --force --types prod,dev,optional --module-dir app",
     "flow": "flow",
     "flow-typed": "rimraf flow-typed/npm && flow-typed install --overwrite || true",
+    "format": "prettier --single-quote --list-different '**/*.js'",
```
amilajack: I think the `"format"` script should be removed and the `"format-fix"` should be merged into the `"lint-fix"` script.

amilajack: This will make it easier for existing users to upgrade without changing their workflow

zhaozhiming:
@amilajack I have done the request change. Please review. Thanks.  
@BuckyMaler @ziedhajsalah Hi, I have resolve the conflicts with your code. Could you review my merge code?  If I broke something, please tell me. Thanks.

package.json
zhaozhiming: done

amilajack:
Thanks! I will review this soon and merge

package.json
```diff
@@ -31,6 +32,12 @@
     "test-watch": "npm test -- --watch"
   },
   "browserslist": "electron 1.6",
+  "lint-staged": {
+    "*.js": [
+      "prettier --single-quote --write",
```
amilajack: Shouldn't `lint-staged` auto fix with eslint as well?

package.json
```diff
@@ -13,14 +13,15 @@
     "flow": "flow",
     "flow-typed": "rimraf flow-typed/npm && flow-typed install --overwrite || true",
     "lint": "cross-env NODE_ENV=development eslint --cache --format=node_modules/eslint-formatter-pretty .",
-    "lint-fix": "npm run lint -- --fix",
+    "lint-fix": "npm run lint -- --fix && prettier --single-quote --write '**/*.js'",
```
amilajack: We're linting `**/*.js` but files in the `**/**/*.js` glob are linted as well. Hmm interesting. For example, `app/components/*.js` files are listed. Just wondering why this is

zhaozhiming: Prettier use [glob syntax from the glob module](https://github.com/isaacs/node-glob/blob/master/README.md#glob-primer). 
`**` matches zero or more directories and subdirectories.

> The following characters have special magic meaning when used in a path portion:
> * ...
> * ** If a "globstar" is alone in a path portion, then it matches zero or more directories and subdirectories searching for matches. It does not crawl symlinked directories.

package.json
zhaozhiming: Yes, it can add `eslint --fix`. 
But I think what `eslint --fix` can do as same as prettier(something like `add semicolon` or `double quotes to single quote`) .
I don't known what error eslint can auto-fix but prettier can't.

This is the [eslint official doc](https://eslint.org/docs/user-guide/command-line-interface#--fix) said, but I don't understand what it mean.

If `eslint --fix` has the same effect as prettier, I think keep only prettier is better.

package.json
zhaozhiming: Maybe put the prettier before eslint --fix is better? What do you think?

amilajack: Well if eslint auto fixes through the prettier eslint config then shouldn't we just need to eslint auto fix?

zhaozhiming: I think `eslint --fix` and prettier have some different. There is a example show the different in this [discuss](https://github.com/prettier/prettier-eslint/issues/101).

Now I think the original way is better.  Use `eslint --fix` first and then use prettier to format fixed code. Please forgive my capricious personality.

```js
"lint-fix": "npm run lint -- --fix && prettier --single-quote --write '**/*.js'",
```

PR #1455, opened 2018-03-05
Added stylus support

BRKsReginaldo:
Hey Guys, i just added stylus support to the project :D


amilajack:
We're trying to actually removing things from the boilerplate at this point because of the existing complexity. There's probably a lot of people that prefer some other css preprocessor over stylus.

PR #1473, opened 2018-03-16
Remove publicPath from webpack.config.renderer.prod

### LINKED ISSUE #1426: Extra ./dist on asset files in prod
Hopefully this will help someone.  Or maybe someone can tell what I was doing wrong to begin with.  I have custom font files that were loading fine in dev.  In prod, however, styles.css was looking for the font files in:

```dist/dist/fontfilename.woff```

Changing this line:

https://github.com/chentsulin/electron-react-boilerplate/blob/master/webpack.config.renderer.prod.js#L25

to 

```'publicPath: '',```

fixed the problem.

nessup:
Webpack was adding `/dist` to the beginning of every URL referenced relatively from `styles.css` or `renderer.prod.js`, causing 404 errors on assets referenced from those bundles.

This fix removes the need for `publicPath` to be forced to the proper value in the config for `ExtractTextPlugin`, thus fixing asset references for all transforms in the renderer prod bundle.

Fixes #1426

### DIFF
--- a/webpack.config.renderer.prod.js
+++ b/webpack.config.renderer.prod.js
@@ -22,7 +22,6 @@ export default merge.smart(baseConfig, {
 
   output: {
     path: path.join(__dirname, 'app/dist'),
-    publicPath: './dist/',
     filename: 'renderer.prod.js'
   },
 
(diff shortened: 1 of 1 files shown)


amilajack:
Thanks for this. I'll start testing this on a variety off apps based on the boilerplate before merging. If i remember correctly, these changes did not work on some of the apps ive worked on before but I'll try this again

PR #1497, opened 2018-04-05
Use module property from dev webpack config in DLL webpack config

### LINKED ISSUE #1468: Further modularizing .dev.dll
Since the `module` property of `webpack.config.renderer.dev.dll` contains the same thing as the `module` property in `webpack.config.renderer.dev`, I'm wondering if there was there a reason for the repetition. If not, maybe we could rewrite `webpack.config.renderer.dev.js` and `webpack.config.renderer.dev.dll.js` sorta like this:

```js
// In webpack.config.renderer.dev.dll...
export default merge.smart(baseConfig, {
  // ...
  module: require('./webpack.config.renderer.dev').module
  // ...

// In webpack.config.renderer.dev...
if (module.parent.filename !== 'webpack.config.renderer.dev.dll.js' && !(fs.existsSync(dll) && fs.existsSync(manifest))) {
```

That's the basic idea. Might look weird or have other unintended side effects, but curious if something like this makes sense or can be done better? Happy to PR with a working version

nessup:
Closes #1468 

To avoid a circular dependency, this relies on a check against the parent module's filename for `webpack.config.renderer.dev.dll.js`. It's expected that developers who rename config files will hopefully grep the codebase for this filename before changing it.

This change also includes this check again when configuring `DllReferencePlugin`, because it is no longer guaranteed that the DLL manifest file exists by the time the renderer config specifies plugins.

I played around with the idea of creating a `webpack-merge` strategy that plucks the value of `module` from the dev webpack config, but it wouldn't help much because, in `webpack.config.renderer.dev.js`, the `DllReferencePlugin` would still try to require the manifest (which would not exist).

Lmk if this should be rebased off a different tag.


amilajack:
I changed the base of this PR to the `v0.13.3`. PR's should be made against this branch instead of master. Can you please fix the merge conflicts.

nessup:
Conflicts resolved

amilajack:
Thanks for this!

PR #1525, opened 2018-04-24
fix "Browser history not working in electron app"

ericpyle:
see https://github.com/chentsulin/electron-react-boilerplate/issues/1178


ericpyle:
this may be superceded by #1184. If so, please close/reject

https://github.com/chentsulin/electron-react-boilerplate/commit/30aca607f25c774e5766fd8aeca48ead29e17a63

PR #1534, opened 2018-05-02
Code example: Exhibit how Flow types don't work in production build

aguynamedben:
⚠️ DO NOT MERGE, example code only

First off, this boilerplate is epic. Thank you for working on it and making it public.

I'm trying to use flow-runtime in my app. I notice that when I build the production package Flow types in classes I've added don't seem to work.

How I'm building the app:
```
yarn
DEBUG_PROD=true yarn package
```

What I see when I run the app:
developer_tools_-_file____applications_electronreact_app_contents_resources_app_asar_app_html

I notice that in .babelrc the flow-runtime plugin is only specified for development. I tried various ways of adding flow-runtime to production, but get other errors about the transform-decorators-legacy plugin.

How can I make sure Flow types work in my own classes in the production builds?

### DIFF
--- a/app/TestClass.js
+++ b/app/TestClass.js
@@ -1,0 +1,19 @@
+// @flow
+
+type GoogleAccount = {
+  upstreamId: string,
+}
+
+export default class TestClass {
+  static accountFlowType = GoogleAccount;
+
+  constructor() {
+    console.log(`Creating instance of TestClass`);
+
+    const account: GoogleAccount = {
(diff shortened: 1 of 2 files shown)


aguynamedben:
I got this working on this branch: https://github.com/chentsulin/electron-react-boilerplate/compare/master...aguynamedben:flow-definitions-in-prod-working?expand=1 Will send a separate proper PR.

PR #1536, opened 2018-05-03
Allow flow definitions to be referenced in production build

aguynamedben:
In certain cases, we may want code running in production to be able to
access Flow type definitions. By default, flow-runtime allows that,
and for performance reasons flow-runtime only runs assertions when
`process.env.NODE_ENV === 'development'`.

This change makes Flow types accessible to code, but assertions just
aren't run. This makes Flow types useable in dynamic interesting
ways. For example, a class could define a property that points to its
Flow type so other code using the class can do additional type
checking given a dynamic Flow type.

Options and default values for babel-plugin-flow-runtime:
- https://codemix.github.io/flow-runtime/#/babel-plugin-flow-runtime
- Note: Assertions only run in dev, no need to re-specify in .babelrc

From: https://codemix.github.io/flow-runtime/#/
>  It does incur a performance hit though, so it's often sensible to
turn off
> assertion mode in production, leaving type declarations intact but
eliminating
> the associated overhead.

NOTE: The updates to yarn.lock are necessary because we need to install `babel-plugin-transform-decorators-legacy` in the non-development area of the app (it's a runtime dependency).

Working example: https://github.com/chentsulin/electron-react-boilerplate/compare/master...aguynamedben:flow-definitions-in-prod-working?expand=1

GitHub Issue: https://github.com/chentsulin/electron-react-boilerplate/issues/1535

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,7 +7,11 @@
     "stage-0",
     "react"
   ],
-  "plugins": ["add-module-exports"],
+  "plugins": [
+    "add-module-exports",
+    "flow-runtime",
+    "transform-decorators-legacy"
+  ],
   "env": {
     "production": {
       "presets": ["react-optimize"],
@@ -16,11 +20,7 @@
     "development": {
       "plugins": [
         "transform-class-properties",
-        "transform-es2015-classes",
-        ["flow-runtime", {
-          "assert": true,
-          "annotate": true
-        }]
+        "transform-es2015-classes"
       ]
     }
   }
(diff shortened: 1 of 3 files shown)


amilajack:
While this makes sense for certain applications, I dont think that it would be a good fit as a default for apps. I think this should be an opt-in feature. I think a good solution would be to document these changes so that anyone interested can make the necessary changes to have flow types in prod. Thoughts on that?

aguynamedben:
@amilajack I think that's totally fair! Hopefully this issue provides a good reference for people that have this question in the future.

PR #1580, opened 2018-06-04
Migrate to mini-css-extract-plugin

BuckyMaler:
Migrate from `extract-text-webpack-plugin` to `mini-css-extract-plugin`.

`webpack.config.renderer.prod.js` doesn't emit a `style.css.map` with these changes. However, there was no source mapping for CSS in production prior to this. The emitted `style.css.map` is empty. If source mapping is intended for CSS in production then an issue should be created.


amilajack:
Awesome! I'll merge this into `v0.14.0` and test this out. Thanks for this 🎉 🎉 🎉

PR #1584, opened 2018-06-08
Move UglifyJSPlugin

BuckyMaler:
`mode: 'production'` is specified in `webpack.config.main.prod.js` and `webpack.config.renderer.prod.js`. That enables `UglifyJSPlugin`. See [default optimization options](https://webpack.js.org/concepts/mode/). Specifying a custom `UglifyJSPlugin` instance in `optimization.minimizer` will [override the default minimizer](https://webpack.js.org/configuration/optimization/#optimization-minimizer).

### DIFF
--- a/webpack.config.main.prod.js
+++ b/webpack.config.main.prod.js
@@ -25,12 +25,16 @@ export default merge.smart(baseConfig, {
     filename: './app/main.prod.js'
   },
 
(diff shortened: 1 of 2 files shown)


amilajack:
Awesome! Thanks for this 🎉

PR #1606, opened 2018-06-26
json linting added

vikr01:


.eslintignore
```diff
@@ -51,3 +51,6 @@ main.js.map
 .idea
 npm-debug.log.*
 __snapshots__
+
+# Package.json
+package.json
```
amilajack: How come `package.json` is ignored?

package.json
```diff
@@ -38,6 +38,10 @@
       "cross-env NODE_ENV=development eslint --cache --format=node_modules/eslint-formatter-pretty",
       "prettier --ignore-path .eslintignore --single-quote --write",
       "git add"
+    ],
+    "*.json|.(babelrc|prettierrc|eslintrc)": [
```
amilajack: What uses this configuration?

PR #1615, opened 2018-06-28
Add electron-redux

allenan:
Due to how electron works, redux is currently only available in the renderer process, which means we can't dispatch redux actions from the main process. For example, we might want to use menu items to dispatch events to our application.

This has been a long-standing issue with this boilerplate:
https://github.com/chentsulin/electron-react-boilerplate/issues/108
https://github.com/chentsulin/electron-react-boilerplate/issues/118
https://github.com/chentsulin/electron-react-boilerplate/issues/968

[electron-redux](https://github.com/hardchor/electron-redux) offers a solution to this problem: run redux in the main process and use IPC to proxy all renderer processes back to the main process.

This PR integrates electron-redux with electron-react-boilerplate and as a demo, allows the counter to be controlled by menu items

image

I used [timesheets](https://github.com/hardchor/timesheets) as a guide for the implementation, though I wasn't 100% sure on the use of history and initialState. So if anyone has recommendations there, that would be helpful!

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,11 +1,14 @@
 import React from 'react';
 import { render } from 'react-dom';
 import { AppContainer } from 'react-hot-loader';
+import { getInitialStateRenderer } from 'electron-redux';
 import Root from './containers/Root';
-import { configureStore, history } from './store/configureStore';
+import { configureStore } from './store/configureStore';
+import history from './store/history';
 import './app.global.css';
 
-const store = configureStore();
+const initialState = getInitialStateRenderer();
+const store = configureStore(initialState, 'renderer');
 
 render(
   <AppContainer>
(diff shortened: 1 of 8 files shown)


Slapbox:
This looks very cool, but is `electron-redux` actively developed or maintained by anyone at this point?

allenan:
`electron-redux` seems to be the best solution out there, but if there's anything more actively-maintained, happy to use that instead

faroutchris:
I merged this into my own project and it works great with having several React apps/split entry points for renderer processes and a shared state between them.

allenan:
I'm using this implementation in one of my projects as well and it works great for me too

jeffvandyke:
Any updates to this? It'd be great to have a standard way to use Redux with multiple windows/dialogs sharing application state.

amilajack:
I think this is a standard feature that should be added to ERB. I think users of the boilerplate would expect this to work out of the box. And it also doesn't add much extra configuration. I will start working on this in another PR. Sorry for taking so long on this 😢

PR #1643, opened 2018-07-20
Update lint-staged for more languages

vikr01:
- Removed unneeded `<div>...</div>` blocks in favor of `<Fragment>...</Fragment>` (should switch to `<>...</>`  in the future).
- Added linting for JSX, CSS, SCSS, Markdown, and YAML to lint-staged

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -42,7 +42,6 @@
     "promise/always-return": "error",
     "promise/catch-or-return": "error",
     "promise/no-native": "off",
-    "quotes": ["error", "single"],
     "react/sort-comp": [
       "error",
       {
@@ -59,13 +58,6 @@
     "react/jsx-filename-extension": [
       "error",
       { "extensions": [".js", ".jsx"] }
-    ],
-    "react/jsx-curly-brace-presence": [
-      "warn",
-      {
-        "props": "always",
-        "children": "ignore"
-      }
     ],
     "react/prefer-stateless-function": "off"
   },
--- a/.eslintrc
+++ b/.eslintrc
@@ -42,6 +42,7 @@
     "promise/always-return": "error",
     "promise/catch-or-return": "error",
     "promise/no-native": "off",
+    "quotes": ["error", "single"],
     "react/sort-comp": [
       "error",
       {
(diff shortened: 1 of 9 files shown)


.vscode/settings.json
```diff
@@ -4,7 +4,23 @@
     ".babelrc": "jsonc",
     ".prettierrc": "jsonc"
   },
+
```
amilajack: Please remove all the terminal changes

app/components/Counter.js
```diff
@@ -25,34 +25,42 @@ export default class Counter extends Component<Props> {
...
     return (
       <div>
-        <div className={styles.backButton} data-tid="backButton">
+        <div className={styles.backButton} data-tid={'backButton'}>
           <Link to={routes.HOME}>
-            <i className="fa fa-arrow-left fa-3x" />
+            <i className={'fa fa-arrow-left fa-3x'} />
           </Link>
         </div>
-        <div className={`counter ${styles.counter}`} data-tid="counter">
```
amilajack: Why are the quotes changed? Is there a particular reason for this change? If not its best to stick with convention

.eslintrc
```diff
@@ -42,6 +42,7 @@
     "promise/always-return": "error",
     "promise/catch-or-return": "error",
     "promise/no-native": "off",
+    "quotes": ["error", "single", { "avoidEscape": true }],
```
amilajack: Please explain why you made this change

.vscode/settings.json
vikr01: Removed.

app/components/Counter.js
vikr01: I believe the rule is off by default, and one convention should be followed. I switched the rule from "always" to "never" now.

.eslintrc
vikr01: Removed, but this was to be able to test since prettier is being used with single quote.

.eslintrc
```diff
@@ -59,6 +59,13 @@
       "error",
       { "extensions": [".js", ".jsx"] }
     ],
+    "react/jsx-curly-brace-presence": [
```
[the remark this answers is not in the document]
vikr01: Removed.

amilajack:
Please rebase this branch against v0.14.0

amilajack:
🎉 🎉

PR #1654, opened 2018-07-26
fixed the outdated redux dev tools documents url

fangzhzh:
- fixed the outdated redux dev tools documents url
    + it moved to http://extension.remotedev.io/docs/API/Arguments.html


amilajack:
Thanks!

PR #1657, opened 2018-07-29
Force fixed local node version

amilajack:


amilajack:
This module is unstable. Will open another PR using a more stable version of this module

PR #1672, opened 2018-08-05
Add route based code splitting

amilajack:


vikr01:
I think you should do
```javascript
() => import('<Component>')
```
instead of
```javascript
import(() => import('<Component>'))
```

Also noticed you tried to import `'./containers//HomePage'` instead of `'./containers/Homepage'`

amilajack:
Yea that was a mistake. Did that locally and it still doesn't work. Still investigating

amilajack:
I think it would be more appropriate to create an example out of this instead of add this to erb. We should document this in erb site under a performance section

PR #1675, opened 2018-08-06
Removed unneeded import, Fixed component file paths

vikr01:


amilajack:
e2e tests are failing

PR #1680, opened 2018-08-06
Linting pattern fixes

vikr01:
- Fixed all linting glob patterns to match zero or more times instead of one time. `.(foo|bar)` would now be ~~`.*(foo|bar)`~~ `.{foo,bar}` ([followed glob patterns](https://github.com/isaacs/node-glob/blob/master/README.md#glob-primer))

~~- Removed .prettierrc in favor of using the `--parser json` flag~~

- Added `stylelint --fix` to lint-staged for css & scss

~~- Added `lint-json-fix` script~~


package.json
```diff
@@ -34,13 +36,22 @@
...
       "git add"
     ],
-    "*.(json|css|scss|md)|.(babelrc|prettierrc|eslintrc)": [
-      "prettier --ignore-path .eslintignore --write",
+    "*.json|.*(babelrc|eslintrc|stylelintrc)": [
+      "prettier --ignore-path .eslintignore --parser json --write",
+      "git add"
+    ],
+    "*.*(css|scss)": [
+      "stylelint --ignore-path .eslintignore --syntax scss --fix",
```
amilajack: How come we're using two css formatters here? Why not one? Does `stylelint` account for cases that `prettier` doesn't?

.prettierrc
```diff
@@ -1,11 +0,0 @@
-{
```
amilajack: Why was this removed?

package.json
```diff
@@ -14,14 +14,16 @@
     "flow-typed": "rimraf flow-typed/npm && flow-typed install --overwrite || true",
     "lint": "cross-env NODE_ENV=development eslint --cache --format=node_modules/eslint-formatter-pretty .",
     "lint-fix": "yarn --silent lint --fix; exit 0",
+    "lint-json-fix": "prettier --ignore-path .eslintignore --parser json --write '**/*(*.json|.*(babelrc|eslintrc|stylelintrc))'",
```
amilajack: Why did we add a new script for this instead of reuse an existing script?

.prettierrc
vikr01: I added this file before to recognize .babelrc, .eslintrc, and .prettierrc as json filetypes. But adding the flag `--parser json` is better.

package.json
vikr01: Add to lint-fix?

package.json
vikr01: `stylelint` caught a missing generic font error, which `prettier` couldn't.
`stylelint` seems to be the equivalent of `eslint` for styles, so I think it's better to keep it.

package.json
vikr01: Could make postlint-fix into this, but it would be long.
`"postlint-fix": "concurrently \"prettier --ignore-path .eslintignore --single-quote --write '**/*.*(js|jsx)'\" \"prettier --ignore-path .eslintignore --parser json --write '**/*(*.json|.*(babelrc|eslintrc|stylelintrc))'\""`

vikr01: Could make postlint-fix into this, but it would be a bit long.
`concurrently \"prettier --ignore-path .eslintignore --single-quote --write '**/*.*(js|jsx)'\" \"prettier --ignore-path .eslintignore --parser json --write '**/*(*.json|.*(babelrc|eslintrc|stylelintrc))'\"`

amilajack: One of the issues with the boilerplate is that is has way too many npm scripts. Try combining this with the `lint-fix` script

vikr01: Merged into `postlint-fix`. This required `.prettierrc` to be added back to avoid the `--parser json` flag (since it would cause errors for js/jsx).

The problem with doing this though, is that if a user wanted to add another filetype to be included in the lint they need to include it in both .prettierrc and package.json.

vikr01: Also used the more commonly used glob patterns, since it reduces the length of the command. To be consistent, I also had to change it in the lint-staged config and `postlint-styles-fix`.

(i.e. `*.*(js|jsx)` becomes `*.{js,jsx}`)

amilajack:
Good stuff. Also can you investigate the flow issue that is causing CI to fail

vikr01:
PR'd  #1712.

Issue #1639, opened 2018-07-18
Doesn't work with yarn workspaces


Vahagn1993:
It works great but I need to use this boilerplate in monorepo project and I faced following error :
```[3/4] Linking dependencies...
[4/4] Building fresh packages...
[-/5] ⠁ waiting...
[-/5] ⠁ waiting...
[5/5] ⠁ electron-react-boilerplate
[-/5] ⠂ waiting...
error D:\V.lab\workspace\Web\workspace\node_modules\electron-react-boilerplate: Command failed.
Exit code: 1
Command: node -r babel-register internals/scripts/CheckNativeDep.js && npm run flow-typed && npm run build-dll && electron-builder install-app-deps && node node_modules/fbjs-scripts/node/check-dev-engines.js package.json
Arguments:
Directory: D:\V.lab\workspace\Web\workspace\node_modules\electron-react-boilerplate
Output:
D:\V.lab\workspace\Web\workspace\node_modules\babel-core\lib\transformation\file\options\option-manager.js:328
        throw e;
        ^

Error: Options {"loose":true} passed to D:\V.lab\workspace\Web\workspace\node_modules\babel-preset-env\lib\index.js which does not accept options. (While processing preset: "D:\\V.lab\\workspace\\Web\\workspace\\node_modules\\babel-preset-env\\lib\\index.js") (While processing preset: "D:\\V.lab\\workspace\\Web\\workspace\\node_modules\\babel-preset-env\\lib\\index.js")
    at D:\V.lab\workspace\Web\workspace\node_modules\babel-core\lib\transformation\file\options\option-manager.js:314:17    at Array.map (<anonymous>)
    at OptionManager.resolvePresets (D:\V.lab\workspace\Web\workspace\node_modules\babel-core\lib\transformation\file\options\option-manager.js:275:20)
    at OptionManager.mergePresets (D:\V.lab\workspace\Web\workspace\node_modules\babel-core\lib\transformation\file\options\option-manager.js:264:10)
    at OptionManager.mergeOptions (D:\V.lab\workspace\Web\workspace\node_modules\babel-core\lib\transformation\file\options\option-manager.js:249:14)
    at OptionManager.init (D:\V.lab\workspace\Web\workspace\node_modules\babel-core\lib\transformation\file\options\option-manager.js:368:12)
    at compile (D:\V.lab\workspace\Web\workspace\node_modules\babel-register\lib\node.js:103:45)
[...]

Issue #1573, opened 2018-05-29
Too many config files


Ellmen:
Including configs for things like flow, vscode, and travis adds a lot of mess to clean up for people who don't use those things. Since this is *the* react electron repo, I feel like it should be stripped down a little.

amilajack:
Which configs should we get rid of?

aguynamedben:
Just registering another opinion... I think the flow configs are good because they provide an opinionated best practice, which is what boilerplate is for.

I think the editor-specific configs could either be removed (because everybody uses a different editor, and your choise editor isn't necessarily related to the project ecosystem), or it should be an explicit policy that other editor configs are welcomed and maybe put them in editor_configs/ with a README file in that directory explaining the policy. I use Emacs, but VS Code seems to be an emerging norm in the JS community so it's probably helpful to have a config laying around.

Also, wouldn't removing yarn.lock make it pretty difficult to debug issues people are having with the project? yarn.lock in projects I've worked on is always committed to make sure that people are using the same dependencies without having to continually ask everybody to list which version they have installed locally.

In general my opinion is:
* It's not that hard to remove config files, and they rarely get in the way. You're going to have to customize this repo for your project anyways. `git rm`ing or editing a few config files you don't find helpful is much easier than being a newbie and not having access to sensible config files. For example, being new to the whole ecosystem it would be very hard to create a Flow config from scratch.
* The purpose of boilerplate is to help people get going in an ecosystem quickly.
* Example config files, even if they're not setup out of the box the way you'd like, are very helpful as it's difficult to come up with from scratch when you're new to an ecosystem.
* Config files are a good opportunity to set opinionated best practices, which is especially useful in boilerplate projects. These best practices reflect norms in a community.

Boilerplate seems like the BEST place for config files. The cost of removing/modifying config to fit your preferences is much lower than the cost of not having ready access to example configs.

Ellmen:
1. I like the idea of an editor_configs and I can appreciate that having sample code makes it easier for a newcomer to get started.

2. The main yarn.lock is certainly important - the one I was talking about is in the app directory and is empty (there are no dependencies in that directory's package.json)

3. I would argue that Flow isn't actually a best practice and integrating it so deeply makes this project less flexible. Skype, GitHub, and Slack are all using TypeScript. The problem with enforcing a certain stack is that when any part of it becomes obsolete, the whole project does.

[...]

PR #1686, opened 2018-08-08
Moved webpack configs to configs folder

vikr01:
#1573 said there were too many configs in the home directory.

Since webpack configs are already being specified from CLI with the `--config` flag, moving them to another folder wouldn't cause any issues.

Also changed any instances of `process.cwd()` to `path.join(__dirname, '..')` so that users using ERB as a package in a monorepo won't face issues with incorrect directories when building. (#1639)


amilajack:
Can you rebase this as well

vikr01:
Rebased against v0.16.0.

PR #1713, opened 2018-08-20
Remove react-router-redux in favor of connceted-react-router

vikr01:
`react-router-redux` is deprecated and no longer maintained. (https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux)

They recommend using `connected-react-router` (https://github.com/supasate/connected-react-router) instead.


amilajack:
Can you rebase this against `v0.16.0`. Will make it easier to review

PR #1716, opened 2018-08-21
Resolve .json & .jsx

vikr01:
In the webpack config, the extensions `.js`, `.jsx`, and `.json` are being resolved ([link](https://github.com/chentsulin/electron-react-boilerplate/blob/2c37e738fd45fd4826256fa62e7876fefe9e36b1/webpack.config.base.js#L62)). We should stay consistent with that in anything being processed by webpack.

- Added the `.jsx` and `.json` extensions to jest
  - Added `.jsx` to be transformed by `babel-jest`
- Removed `.json` from any imports in the `./app` directory (since `./app` is what we're processing through webpack)
- Added eslint rule to enforce omitting `.json` from imports within the `./app` directory only


amilajack:
Can you rebase this against `v0.15.0`. Will make it easier for me to review

vikr01:
Rebased against `v0.15.0`. Travis CI build is failing from `yarn flow` due to the types inside `store` that was fixed in master.

app/.eslintrc
```diff
@@ -0,0 +1,9 @@
+{
```
amilajack: Why not add this to root `.eslintrc`?

vikr01:
Didn't realize ES resolves `json`.

Applied the rule to the root eslintrc and resolved all `json` imports.

PR #1722, opened 2018-08-25
Watch change of electron-main-process code and reload in dev env

HyunmoAhn:
## What I want?

I think that this repository is a great guide to the start of a project using electron and react.

Currently, modification of the code for the main process of electron is not automatically reload in the development environment. For example, you add code to print something like `console.log('something')` in `app/main.dev.js` then you must close process and rerun `npm run dev` to show print 'something'. 

This repository have tiny code about main process of electron(`app/main.dev.js`, `app/menu.js`). So the function to automatically reload may not be necessary. But I hope it will help other developer who refer to this repository. 

## Did this PR

Using [nodemon](https://github.com/remy/nodemon), watch change of `js` code and rerun about starting electron. 

nodemon watch only `app/*.js` code. 
If watch all about js code, The react-hot-load is useless because every change occur to rerun app. It is not efficient. 
There exist unrelated to main process of electron In `app/*.js` path like `index.js` and `Routes.js`.
Current nodemon setting occur reload when it is changed. It is obviously misused. But It is rarely modified so I don't think it's a big problem.
To fix, To make directory about main process of electron is solution like that.
```
app/main/         - make new directory "main"
ㄴ main.dev.js
ㄴ menu.dev.js
.... another main process code.

package.json

"start-main-dev-watch": "nodemon --exec \"npm run start-main-dev\" --watch app/main",
```
But I think it doesn't need because this repository have tiny code about main process of electron.

## Usage
```
npm run dev
```
After edit `app/main.dev.js` code.
It automatically reload!


amilajack:
While the idea is good, this isn't the best solution for the boilerplate. We should take advantage of webpack for the main process in development. We'll need a webpack config for the main process like so:
https://github.com/amilajack/erb-local-fs-read-example/blob/master/webpack.config.main.dev.js

PR #1730, opened 2018-08-28
Fix issues created by adding connected-react-router

vikr01:
`react-router-redux` was still being used in some areas, switched that for `connected-react-router`.

`routerReducer` is not currently exposed by `connected-react-router`, so I followed a "hack" mentioned [here](https://github.com/supasate/connected-react-router/issues/79#issuecomment-403271755) to add that back.


amilajack:
After https://github.com/supasate/connected-react-router/pull/132 gets merged let's remove the hack. Thanks for the PRs!

Issue #1689, opened 2018-08-10
Monorepo support


vikr01:
Using ERB in a subdirectory will cause problems when its `node_modules` are hoisted. (#1639) The current workaround is to [avoid hoisting its dependencies](https://yarnpkg.com/blog/2018/02/15/nohoist/), but this isn't ideal in the long run.

Areas causing problems:
- [x] `CheckNativeDep.js` checking node_modules in [current working directory](https://github.com/chentsulin/electron-react-boilerplate/blob/master/internals/scripts/CheckNativeDep.js#L12)
  - Fixed in #1854
- [x] Certain scripts in `package.json` expecting `node_modules` to be in the current working directory (i.e. [`./node_modules/webpack/bin/webpack`](https://github.com/chentsulin/electron-react-boilerplate/blob/master/package.json#L8)).
  - Fixed build scripts in #1827 
  - Fixed lint script in #1746

Possible solutions:
- `webpack.config.*.babel.js`
- `require.resolve`
- `npm root`

rntjr:
project supports monorepo in release 1.3.0?

PR #1739, opened 2018-08-31
WIP: Use node-modules-path module to get path to node_modules

vikr01:
Partially helps #1689.

Avoid assuming the location of `node_modules`. Instead, use `node-modules-path` to get the path to the nearest `node_modules`.

We will also have to switch the jest config in `package.json` to a `jest.config.js` to utilize `node-modules-path` for `moduleDirectories`. So for now, the assumption is still there during testing.

### DIFF
--- a/app/main.dev.js
+++ b/app/main.dev.js
@@ -11,6 +11,7 @@
  * @flow
  */
 import { app, BrowserWindow } from 'electron';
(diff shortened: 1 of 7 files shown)


vikr01:
This won't be enough, unfortunately.

Issue #1734, opened 2018-08-29
Migrate to babel@7


amilajack:

vikr01:
Been working on it at [/vikr01/electron-react-boilerplate/tree/feature/babel-7](/vikr01/electron-react-boilerplate/tree/feature/babel-7).

When running commands from the root directory, everything works fine. When running from the `app` directory, babel is not finding its config.
Issue seems to be related to babel believing the project is a monorepo due to `app/package.json`. [Switched `.babelrc` to `babel.config.js`](https://babeljs.io/docs/en/config-files#project-wide-configuration) and still no luck.

PR #1749, opened 2018-09-04
chore(deps): bump deps

amilajack:


vikr01:
`babel-loader` 8 is intended for `babel` 7. We'll update that when we resolve #1734.

amilajack:
Oh right. Forgot to ignore all the babel dep updates

PR #1755, opened 2018-09-05
Runtime optimizations

vikr01:
Using `dependencyKeys = Object.keys(dependencies)`then later using `dependencyKeys.includes` creates an unnecessary linear runtime. Simply using `dependencies` as a map is enough. `dependencyKeys` is not used for anything else so it can thus be removed.

We also don't need the whole package tree when running `npm ls` since we only care about the root dependencies -- so we can just use `--depth 0` to grab only those root dependencies.

### DIFF
--- a/internals/scripts/CheckNativeDep.js
+++ b/internals/scripts/CheckNativeDep.js
@@ -7,7 +7,6 @@ import { dependencies } from '../../package';
 (() => {
   if (!dependencies) return;
 
-  const dependenciesKeys = Object.keys(dependencies);
   const nativeDeps = fs
     .readdirSync('node_modules')
     .filter(folder => fs.existsSync(`node_modules/${folder}/binding.gyp`));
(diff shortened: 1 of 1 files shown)


amilajack:
We need to check the entire dependency tree for the native deps. module a could require module which requires a native dep

vikr01:
Here's an example of what I mean:

```bash
$ npm ls fbjs --json
electron-react-boilerplate@0.16.0 <path-to-erb>
{ ...
  "dependencies": {
    "enzyme-adapter-react-16": {
      "version": "1.1.1",
      "dependencies": {
        "react-reconciler": {
          "version": "0.7.0",
          "dependencies": {
            "fbjs": {
              "version": "0.8.17"
            }
          }
        }
      }
    },
    "react": {
      "version": "16.4.1",
      "dependencies": {
        "fbjs": {
          "version": "0.8.17"
        }
      }
    },
    "react-dom": {
      "version": "16.4.1",
      "dependencies": {
        "fbjs": {
          "version": "0.8.17"
        }
      }
    },
    "react-test-renderer": {
      "version": "16.4.1",
      "dependencies": {
        "fbjs": {
          "version": "0.8.17"
        }
      }
    }
  }
}
```

```bash
$ npm ls fbjs --depth 0 --json
{ ...
  "dependencies": {
    "enzyme-adapter-react-16": {
      "version": "1.1.1"
    },
    "react": {
      "version": "16.4.1"
    },
    "react-dom": {
      "version": "16.4.1"
    },
    "react-test-renderer": {
      "version": "16.4.1"
    }
  }
}
```

Notice that the keys do not change, only the children of them.
Since [`Object.keys` is the only operation being run on this object](https://github.com/chentsulin/electron-react-boilerplate/blob/b57f77cf54ea67143102af982f3e662b9e88dd2c/internals/scripts/CheckNativeDep.js#L22), we don't need the _whole_ tree -- just the top level.

amilajack:
This can be closed since the `CheckNativeDep` script is no longer used

PR #1759, opened 2018-09-06
Use npm-run-all to shorten commands

vikr01:
With `npm-run-all`, we can avoid unnecessarily long scripts in `package.json`. For example, `yarn foo && yarn bar` becomes `npm-run-all foo bar`.

There is also a `--parallel` flag, which we can use instead of `concurrently`.

Note that while the name is `npm-run-all`, [it is compatible with `yarn`](/mysticatea/npm-run-all#yarn-compatibility).


amilajack:
Another direction would be to 'use-the-platform' and use `&` for concurrency. ex:
```bash
$ yarn build-main & yarn build-renderer
```
Forgot why we chose to use `concurrently`. I'll investigate this more

PR #1890, opened 2018-10-29
Remove app/node_modules from .gitignore

aguynamedben:
In previous versions of electron-react-boilerplate, there was a
2-package.json structure where modules were installed in
./node_modules and ./app/node_modules. We've since migrated to a
single package.json system, and there should only be ./node_modules
present. If you notice ./app/node_modules (i.e. when running `git
status`), you can safely delete the director ./app/node_modules. All
dependencies are downloaded in ./node_modules now.

History: https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/1854

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,6 @@ build/Release
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 node_modules
-app/node_modules
 
 # OSX
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,7 @@ build/Release
 
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
-node_modules
+/node_modules
 
 # OSX
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,7 @@ build/Release
 
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
-/node_modules
+node_modules
 
 # OSX
 .DS_Store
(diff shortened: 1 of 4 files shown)


aguynamedben:
Looks like the build is failing because app/node_modules is also referenced in .travis.yml

aguynamedben:
Actually the failure is `Error: Please specify author 'email' in the application package.json`, seems unrelated? Let me know if you'd like me to remove app/node_modules from .travis.yml, seems to me like it should be removed there as well.

vikr01:
@aguynamedben Yes, that should be removed from `.travis.yml`.

.gitignore
```diff
@@ -25,8 +25,7 @@ build/Release
 
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
-node_modules
-app/node_modules
+/node_modules
```
amilajack: Forcing `node_modules` to only be ignored in the root feels like too strict of a config. Thoughts on that? Wht about changing this should be changed to `node_modules`?

aguynamedben: @amilajack That was a response to this comment from @vikr01: https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/1854#issuecomment-434074811

I could go either way on this. I can't think of any use case where a user *would* have node_modules in a subdirectory, but maybe?

vikr01: Yeah, I was thinking the same thing (conversation on [#1854](https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/1854#issuecomment-434074811)).

This actually could cause people using `electron-react-boilerplate` with subpackages to all of a sudden see their `node_modules` be included in unstaged files.

aguynamedben: Yeah, good point. Sounds like I should change it back to just node_modules... incoming...

amilajack:
Thanks for this!

PR #1895, opened 2018-10-29
babel target current electron version

amilajack:


vikr01:
Once we upgrade to `babel` 7, I'd rather we use `require('electron/package.json').version`.

amilajack:
`"electron": "current"` would be preferred but this doesn't work yet. I think we should PR preset env config to add that

PR #1921, opened 2018-10-31
[WIP]: add initial electron-redux support

amilajack:
Closes https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/1920

- [x] Add initial implementation
- [ ] Add example of usage

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,11 +1,13 @@
 import React from 'react';
 import { render } from 'react-dom';
 import { AppContainer } from 'react-hot-loader';
+import { getInitialStateRenderer } from 'electron-redux';
 import Root from './containers/Root';
 import { configureStore, history } from './store/configureStore';
 import './app.global.css';
 
-const store = configureStore();
+const initialState = getInitialStateRenderer();
+const store = configureStore(initialState, 'renderer');
 
 render(
   <AppContainer>
--- a/app/store/configureStore.dev.js
+++ b/app/store/configureStore.dev.js
@@ -3,6 +3,13 @@ import thunk from 'redux-thunk';
 import { createHashHistory } from 'history';
 import { routerMiddleware, routerActions } from 'connected-react-router';
 import { createLogger } from 'redux-logger';
+import {
+  forwardToMain,
+  forwardToRenderer,
+  triggerAlias,
+  replayActionMain,
+  replayActionRenderer
+} from 'electron-redux';
 import createRootReducer from '../reducers';
 import * as counterActions from '../actions/counter';
 import type { counterStateType } from '../reducers/types';
@@ -11,9 +18,12 @@ const history = createHashHistory();
 
 const rootReducer = createRootReducer(history);
 
-const configureStore = (initialState?: counterStateType) => {
+const configureStore = (
+  initialState?: counterStateType,
+  scope: string = 'main'
+) => {
   // Redux Configuration
-  const middleware = [];
+  let middleware = [];
   const enhancers = [];
 
   // Thunk Middleware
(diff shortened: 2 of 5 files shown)


vandosant:
Is this still active? I tried implementing your changes in my application, but I'm getting `Could not find reduxState global in main process, did you forget to call replayActionMain?` -- hoping to see an example usage! Thanks.

amilajack:
Haven't had time to finish it

amilajack:
Haven't spent much time on this because I didn't know if there would be any demand for it. I'll take another look soon. Sorry for the delay on this. Currently busy maintaining other projects ATM

Slapbox:
@amilajack you're one of the busiest folks I know on GitHub, and I and others appreciate all you do! I'm pretty sure the issue here is as simple as changing this line (though I've not tested it outside my own project):

https://github.com/electron-react-boilerplate/electron-react-boilerplate/blob/0b681c3adff664d7a694752ab3fb803be15104f0/app/store/configureStore.dev.js#L59

The issue is that a property of `window` is being checked before we check if it's the `renderer` process/before we check if `window` is defined.

```js
const composeEnhancers = scope === 'renderer' && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
```

amilajack:
@Slapbox I see. So should we move that line to be before https://github.com/electron-react-boilerplate/electron-react-boilerplate/blob/0b681c3adff664d7a694752ab3fb803be15104f0/app/store/configureStore.dev.js#L43?

Slapbox:
@amilajack I think if we do that then `composeEnhancers` will not be defined, as opposed to toggling between `__REDUX_DEVTOOLS_EXTENSION_COMPOSE__` and Redux's built in `compose`.

I'm not sure of any way to submit a PR on a PR, so here's a gist with more or less what I did: https://gist.github.com/Slapbox/a169953ad477cde04a6be53844b0a0dd#file-js-L59

----

Unrelated to getting it working, once this is added it's important we note the importance of having a `payload` property in your actions. Despite not being required in Flux Standard Actions, Electron-Redux will not relay actions missing a `payload` property.

amilajack:
@Slapbox you can PR to my branch. I think that would be the best way to suggest changes. That way we can also check if CI passes when you PR to my branch

NclsL:
A bit late, but I implemented the changes above and the issue stayed. Shouldn't we do
```
import { configureStore } from './store/configureStore';
const store = configureStore(undefined, 'main');
```

in `app/main.dev.js`

or something similar? 
I tried adding those and it only led to another issue that I was unable to solve:

```
A JavaScript error occurred in the main process
Uncaught Exception:
Invariant Violation: Hash history needs a DOM
    at invariant (/home/x/Projects/electron-react-boilerplate/node_modules/invariant/invariant.js:40:15)
    at createHashHistory (/home/x/Projects/electron-react-boilerplate/node_modules/history/createHashHistory.js:69:27)
    at Object.<anonymous> (/home/x/Projects/electron-react-boilerplate/app/store/configureStore.dev.js:17:17)
    at Object.<anonymous> (/home/x/Projects/electron-react-boilerplate/app/store/configureStore.dev.js:122:3)
    at Module._compile (module.js:642:30)
    at Module._compile (/home/x/Projects/electron-react-boilerplate/node_modules/pirates/lib/index.js:83:24)
    at Module._extensions..js (module.js:653:10)
    at Object.newLoader [as .js] (/home/x/Projects/electron-react-boilerplate/node_modules/pirates/lib/index.js:88:7)
    at Module.load (module.js:561:32)
    at tryModuleLoad (module.js:504:12)
```

Slapbox:
I got this working in my project, but I don't use routing. I originally got stuck in the same place as @NclsL when I included routing. Does anyone have any thoughts on working around that? It's never going to make it into the master branch if we can't get it working with routing.

Issue #1957, opened 2018-11-03
`yarn dev` does not open any windows - Windows 10


theamazingfedex:
I have a clean clone of the repository, and after running `yarn` to install all dependencies, I attempt to run `yarn dev`, and it executes successfully, however it hangs at the following point without opening an electron window:
``` sh
$ yarn dev
yarn run v1.12.1
$ cross-env START_HOT=1 node -r @babel/register ./internals/scripts/CheckPortInUse.js && cross-env START_HOT=1 yarn start-renderer-dev
$ cross-env NODE_ENV=development webpack-dev-server --config configs/webpack.config.renderer.dev.babel.js
Starting Main Process...

> electron-react-boilerplate@0.17.0 start-main-dev C:\Users\daniwood\oss\game-discovery-electron
> cross-env HOT=1 NODE_ENV=development electron -r @babel/register ./app/main.dev.js
```

## Prerequisites

- [x] Using yarn
- [x] Using an up-to-date master branch
- [?] Using latest version of devtools. See [wiki for howto update](https://github.com/electron-react-boilerplate/electron-react-boilerplate/wiki/DevTools)
- [ ] Link to stacktrace in a Gist (for bugs)
- [ ] For issue in production release, devtools output of `DEBUG_PROD=true yarn build && yarn start`

## Expected Behavior
I'm expecting the electron window to open.

## Current Behavior
The electron window does not open.

## Steps to Reproduce (for bugs)

1. be on windows 10
2. clone the repository
3. `yarn` to install dependencies
4. `yarn dev` to start the app
5. wait forever as the app never opens

## Context

I have been unable to play around with an electron app, and I just wanted to get in and play around with it.

No.

## Your Environment

- Node version : 11.1.0
- Version or Branch used : current master
- Operating System and version : Windows 10 Enterprise
- Link to your project : this repo

vikr01:
I haven’t tested with Node 11 since it recently came out. Perhaps try with Node 10?

yeluoqiuzhi:
I have the same problem on osx, and I use node v10.13.0

Cethy:
This is silly.
Had the same problem (on windows 8), and the fix was to edit the `package.json` :
`"dev": "cross-env START_HOT=1 node -r @babel/register ./internals/scripts/CheckPortInUse.js && cross-env START_HOT=1 yarn start-renderer-dev",`
to
[...]

PR #1958, opened 2018-11-04
Fix versions in package.json

vikr01:
Project is not currently working with Node 11. [#1957](https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/1957#issuecomment-435618180)

Change `devEngines` to `engines` so `yarn` will throw an error if incorrect versions are used.


amilajack:
We will need to remove this restriction later when `node@11` is stable

Issue #1961, opened 2018-11-04
Bundle size significantly increased in latest version


gorilla-devs:
As of [this commit](https://github.com/electron-react-boilerplate/electron-react-boilerplate/commit/9f4381aeaed82acfff2e1cd59f781bcf5669f595), the app.asar on windows weights 6mb. In the latest commit it weights 46mb. Since my application is much bigger, the size went from 16mb to 122mb, quite an increase. I noticed that now it bundles the node_modules folder inside the app.asar, it didn't do that before. Any idea on how to solve this?

To reproduce this it's enough to do yarn and yarn package on both commits and go look inside release/win-unpacked/resources/app.asar, extract it with [this](https://github.com/electron/asar) et voila

aguynamedben:
The commit you point to is a simple README change... did you mean to paste another commit?

gorilla-devs:
I know. That is the last commit with a small bundle size. The commits after that ([from this commit](https://github.com/electron-react-boilerplate/electron-react-boilerplate/commit/d5416958865bca241f001807cb5e1f62a0cd1ba2)) weights a lot. I need to fix this asap

vikr01:
This sounds like it’s related to the webpack config changes.

thecorsair:
same issue. 
In the last version(0.16.0), the dist folder size is very small and not contains [node_modules].

thecorsair:
> same issue.
> In the last version(0.16.0), the dist folder size is very small and not contains [node_modules].

It's not like 0.16.0, maybe 0.15.0

gorilla-devs:
Anyone got a fix? I tried but i couldn't figure out the problem

vikr01:
@gorilla-devs If you discard the `node_modules` folder entirely after building, does your app still run fine? If that's the case, that may actually be better behavior since your bundle size should be smaller than `node_modules` + your bundled app due to `webpack`'s tree-shaking.

gorilla-devs:
I just tried unpacking app.asar using the asar utility, deleting node_modules and repacking it. It crashes saying it cannot find one of the modules (electron-log). There is surely something wrong in the configs of webpack

gorilla-devs:
If anyone needs, our fantastic @amilajack found a solution (even if it's temporary afaik).
[#1986](https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/1986)

PR #1986, opened 2018-11-09
fix package size bug

amilajack:
All the dependencies were being treated as externals. Because webpack bundles all our node modules, we do not need to include them in our packaged electron app. `electron-builder` adds all files in `**/*` glob to packaged app so this was why `node_modules` was included. This fix will pose issues for projects that uses native deps because webpack cannot handle native deps yet. With our two `package.json` architecture, we did ship `app/node_modules` which was the only solution. We need to investigate solutions for native deps with this single `package.json` architecture.

### DIFF
--- a/configs/webpack.config.base.js
+++ b/configs/webpack.config.base.js
@@ -4,11 +4,8 @@
 
 import path from 'path';
 import webpack from 'webpack';
-import { dependencies } from '../package.json';
 
 export default {
-  externals: [...Object.keys(dependencies || {})],
-
   module: {
     rules: [
       {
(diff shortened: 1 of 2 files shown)


gorilla-devs:
Works! Thanks!

talha131:
> We need to investigate solutions for native deps with this single package.json architecture.

Issue https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/1961 is high priority if not critical. I suspect it would be show stopper for some users. Judging by your quote above, it would be sometime before a good solution is found.

If that's the case then wouldn't it be better if 0.17.0 is marked as pre-release or beta? The transition from two to one package structure does not appear to be smooth.

Issue #1967, opened 2018-11-05
Yarn run start broken in last version


clem91:
## Prerequisites

- [x] Using yarn
- [x] Using an up-to-date master branch
- [x] Using latest version of devtools. See [wiki for howto update](https://github.com/electron-react-boilerplate/electron-react-boilerplate/wiki/DevTools)
- [ ] Link to stacktrace in a Gist (for bugs)
- [ ] For issue in production release, devtools output of `DEBUG_PROD=true yarn build && yarn start`

## Expected Behavior

yarn run start should run the application in production mode without packaging it

## Current Behavior

electron can't read import instruction in index.js

## Possible Solution

Maybe add some transpiling before ?

## Steps to Reproduce (for bugs)
1.yarn run start or yarn start

2. 💔

damianfijorek:
I have the same issue. Restoring the `package.json` with `"main": "./main.prod.js"` in `app` fixed that. It seems that the `"main": "./app/main.prod.js"` in root `package.json` is disregarded.

amilajack:
Interesting. I cannot reproduce this with a fresh clone

vikr01:
For future reference -- please note that if you don't provide a stacktrace or a repo to reproduce then someone has to attempt to reproduce the issue themselves, which means your issue may take some time to get to (and could be closed if it's not being occurring for anyone else).

I tested this on a fresh clone and got the same bug, here's the stacktrace:

```shell
$ cross-env NODE_ENV=production electron ./app/
App threw an error during load
<path-to-boilerplate>/app/index.js:1
(function (exports, require, module, __filename, __dirname, process, global, Buffer) { return function (exports, require, module, __filename, __dirname) { import React from 'react';
                                                                                                                                                           ^^^^^^

SyntaxError: Unexpected token import
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:606:28)
    at Object.Module._extensions..js (module.js:653:10)
    at Module.load (module.js:561:32)
    at tryModuleLoad (module.js:504:12)
    at Function.Module._load (module.js:496:3)
    at loadApplicationPackage (<path-to-boilerplate>/node_modules/electron/dist/Electron.app/Contents/Resources/default_app.asar/main.js:287:12)
    at Object.<anonymous> (<path-to-boilerplate>/node_modules/electron/dist/Electron.app/Contents/Resources/default_app.asar/main.js:328:5)
[...]

PR #1991, opened 2018-11-10
fix incorrect path to electron main prod script in yarn start script

amilajack:
Closes https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/1967

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "postlint-styles-fix": "prettier --ignore-path .eslintignore --single-quote --write '**/*.{css,scss}'",
     "precommit": "lint-staged",
     "prestart": "yarn build",
-    "start": "cross-env NODE_ENV=production electron ./app/",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "postlint-styles-fix": "prettier --ignore-path .eslintignore --single-quote --write '**/*.{css,scss}'",
     "precommit": "lint-staged",
     "prestart": "yarn build",
-    "start": "cross-env NODE_ENV=production electron ./app/",
+    "start": "cross-env NODE_ENV=production electron ./app/main.prod.js",
```
vikr01: Worth noting, `./` can also be used instead, since it'll search the `main` specified in this `package.json`.

amilajack: Yup, that's right. I added it like so because I think most users of the boilerplate are usually new to node and probably wouldn't know that. I thought this was more explicit. Thoughts?

vikr01: Sounds good to me.

PR #2003, opened 2018-11-15
add back electron-rebuild

amilajack:
I should have caught that this was removed when reviewing https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/1854. This should not have been removed. Removing it broke native modules


aguynamedben:
Good catch, I didn't see that either. Are you planning on reverting that PR entirely? (https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/2005) For what it's worth, I have a project that incorporates the single package.json structure and native modules are working well with it (I custom-compile node-sqlite3 with SQLCipher).

amilajack:
Added a new comment here: https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/2005#issue-231285826

PR #2005, opened 2018-11-15
revert "single package.json structure" changes

amilajack:
This reverts https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/1854

## Reasons for Reverting
* This PR made native modules work but it did so at the cost of including all modules, not just native ones, in the build. Shipping the `node_modules` significantly increased the build size of apps for some users.
* Before this PR, we were minimizing what were imported as webpack externals. Dependencies that are labeled as externals are not processed by webpack and are therefore not optimized. These dependencies are at runtime instead of compile time. Webpack does not work well with native modules so usually native modules are treated as externals. In https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/1854, all dependencies were treated as externals. This did make things work for a single package json structure but it came at the cost of performance and app size. For anyone who wants further details feel free to ask


aguynamedben:
I see, thanks for explaining. So going forward, is this the general idea to explain "Why do we have 2-package.json structure":
* We use a 2-package.json structure in order to differentiate run time dependencies (shipped with the packaged app) from compile time dependencies (not shipped with the packaged app).
* If you're using a native module (i.e. it has C++ in it and needs to be rebuilt), it's probably a run time dependency and should go in app/package.json.
* If you're installing a dev tool or a tool that Webpack will compile and optimize, it's probably a compile time dependency and should go in package.json.

Is there any more resources you can offer to further explain this: "Webpack does not work well with native modules so usually native modules are treated as externals." I've run into so many frustrating issue with Webpack+Electron (i.e. require() not working for (electron-remote)[https://github.com/electron-userland/electron-remote]) and would like to learn more.

Thanks for your hard work on this project.

@HyperSprite I'm also curious if you have any ideas around this?

HyperSprite:
I did that for simplicity, another option would be to add an "externals" stanza to the package.json and set them specifically. It means that people need to remember to add them there, something like:
```json
  "externals": [
    "sqlite3",
  ],
```
for instance
and then in the webpack config doing
```js
// Find all the dependencies without a `main` property and add them as webpack externals
function filterDepWithoutEntryPoints(dep: string): boolean {
  // Return true if we want to add a dependency to externals
  try {
    // If the root of the dependency has an index.js, return true
    if (fs.existsSync(path.join(__dirname, `node_modules/${dep}/index.js`))) {
      return false;
    }
    const pgkString = fs
      .readFileSync(path.join(__dirname, `node_modules/${dep}/package.json`))
      .toString();
    const pkg = JSON.parse(pgkString);
    const fields = ['main', 'module', 'jsnext:main', 'browser'];
    return !fields.some(field => field in pkg);
  } catch (e) {
    console.log(e);
    return true;
  }
}

export default {
  externals: [
    ...externals,
    ...Object.keys(possibleExternals || {}).filter(filterDepWithoutEntryPoints),
  ],
```

amilajack:
I dont think that would be a good solution for most of the users of the boilerplate. Many people do not know if they are using native modules or not. Also correct me if i'm wrong but I dont think this addresses the issue of package size because electron-builder would still be including the `./node_modules` in the packaged app

aguynamedben:
@HyperSprite That sounds like a very good solution to me. It's way less complicated to explain that all the confusion around running `yarn` in 2 places, having 2 package.jsons, etc. I.e. the ERB docs can say:

> If you have run time dependencies that should be shipped as-is along with your packaged app, and not optimized by webpack, add them to the "externals" config of package.json. If you have native modules (written in C++, need to be rebuilt), you probably want to ship those as-is as externals, as webpack is not good at packaging and optimizing native modules.

That seems like solution that is much easier to manage long-term and easy to understand as a user.

amilajack:
I couldn't agree more. I really think we should get rid of the two package structure. Eventually I hope the infra around native modules + electron + webpack matures and solves these kinds of issues.

While it is simpler and much more intuitive to use a single package.json structure, we can't go in that direction right now because we dont have the right infra at the moment. I dont think we should sacrifice performance for simplicity. The main focus of ERB is to create a foundation for scalable apps. So many electron apps suffer from having large app sizes and terrible performance. If we leverage webpack to the fullest extent possible then we can circumvent those issues and create better quality apps.

I will think about some ideas for reverting to a single package structure in a way that won't force us to make sacrifices and create a RFC if I come up with something.

HyperSprite:
although, all that said, I don't know how this plays with the _dependencies without a main_ situation since I'm not using those. Those would have to be managed so webpack could build them but they don't need to be included in the final asar either since they should already be packed up with the app. hmm

vikr01:
@HyperSprite That's a good point. I usually make `react` and other dependencies for my web applications as `devDependencies` because they get bundled in with `webpack` ([example](https://github.com/vikr01/tbd-project-name/blob/master/packages/frontend/package.json#L28)). In environments like Heroku, [`devDependencies` get pruned](https://github.com/heroku/heroku-buildpack-nodejs/blob/f6d55d632a84bc1a83d20bb15ce4d186846cff6d/lib/dependencies.sh#L125) because they shouldn't be needed once the app is built.

I really don't like assumptions of where `node_modules` are, which is what the 2-package.json structure does. I really hope we can find a way around that.

aguynamedben:
@HyperSprite I agree that is a really good point. To be clear, I'm not a straight out webpack hater... webpack is really good at bundling everything together at compile time (therefore allowing us to move things from dependencies to devDependencies). I'm going to pursue @HyperSprite's solution for my app too. Those results are pretty exciting.

It seems that moving most packages to devDependencies, except for packages that need to be shipped with the app (i.e. native modules) is a very clear solution. It's even easier to document and explain to users than having them list navite modules in externals.

@amilajack's The issues you mention (large app size, terrible performance) are very important and should be a primary consideration, even if it's a little more complicated. Is there something that causes @HyperSprite's solution to *not* address those sufficiently? Does webpack somehow optimize externals to result in a smaller packaged app or faster loading? It seems like native modules must be shipped as-is with your packaged app, no matter how you slice it (i.e. as webpack externals or via the devDependencies/dependencies approach).

🙏Thank you to everybody for the quality discussion here. I'm learning a lot and these conversations can hopefully lead to clear, simple best practices.

HyperSprite:
@vikr01 yeah, having React in dependencies thing, I think that comes from the SSR side of things where stuff is being compiled at runtime or something. Server side, nobody really cares if you have another 40mb of stuff. Then that pattern was continued in places like CRA and now I think people just do it out of habit (I know I do). It didn't help that eslint Airbnb by default complains if you have imports that are not in dependencies.

I second @aguynamedben comment, this is a great discussion, I'm learning a lot as well.

HyperSprite:
Just in case someone finds this later...
So all of my non-native node_moduels are now in devDependencies and my webpack.config.base externals looks like this
```js
externals: [
    ...Object.keys(dependencies || {}),
    ...Object.keys(devDependencies || {}).filter(filterDepWithoutEntryPoints),
  ],
```
I have another app that has no native deps and I rebuilt it with this setup and no other changes, and there are no `node_modules` in the app.asar and these are the size results:

#### Before:

44719116 Oct 29 18:31  quick-data-inspector_1.0.5_amd64.deb
70592540 Oct 29 18:30  quick-data-inspector-1.0.5-x86_64.AppImage
49519004 Oct 29 18:30 'Quick Data Inspector Setup 1.0.5.exe'

#### After

38039366 Nov 15 21:42  quick-data-inspector_1.0.5_amd64.deb
54512442 Nov 15 21:41  quick-data-inspector-1.0.5-x86_64.AppImage
40060611 Nov 15 21:41 'Quick Data Inspector Setup 1.0.5.exe'

amilajack:
@HyperSprite how many dependencies do you have in your `dependencies`? Also see this issue for about the package sizes https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/1961

vikr01:
@HyperSprite The `no-extraneous-dependencies` rule for [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import) is good for node-based applications since those dependencies are needed at runtime. If we were to go down this road of fixing the deps to devDeps (because they *should* be a `devDependency` if they aren't needed at runtime), then we'd be mixing runtime dependencies and bundled dependencies in our imports -- so the rule may need to be updated so that we can specify what dependencies are getting bundled (and are therefore not extraneous).

HyperSprite:
@vikr01 with this `import/no-extraneous-dependencies: [error, { devDependencies: true }]` it will only complain if the import is not in either devDependencies or dependencies. Right now this repo has the rule is just set to "off" (but I don't know why).

@amilajack I don't have any node_modules in my quick-data-inspector app (and none are being bundled). 
I just unpacked my running asar and noticed it has source maps which I thought were turned off for prod builds but there they are:

190302 Nov 16 14:19 main.prod.js
737334 Nov 16 14:19 main.prod.js.map
864368 Nov 16 14:19 dist/main.prod.js
3138423 Nov 16 14:19 dist/main.prod.js.map
3754180 Nov 16 14:19 dist/renderer.prod.js
8219434 Nov 16 14:19 dist/renderer.prod.js.map
8968 Nov 16 14:19 dist/style.css
23396 Nov 16 14:19 dist/style.css.map

In my app there are 17mb of maps.

HyperSprite:
This is after I deleted the dist dir and repackaged 

The dist dir is now just:
```bash
3754180 Nov 16 14:59 renderer.prod.js
   8968 Nov 16 14:59 style.css
  23396 Nov 16 14:59 style.css.map  # not sure what is causing this
```
These are smaller but still seem pretty big for what is left in the app.asar

```bash
35699340 Nov 16 15:00  quick-data-inspector_1.0.6_amd64.deb
50285076 Nov 16 14:59  quick-data-inspector-1.0.6-x86_64.AppImage
36733493 Nov 16 14:59 'Quick Data Inspector Setup 1.0.6.exe'
```

I'm not really sure when the extra files showed up but I'm thinking during a debug build and then they just never left. I am going to add `CleanWebpackPlugin` to my projects and just wipe out the dist on the start of prod build.

**the webpack dll in the room** 
All of that said, one thing that would also need to be considered if all the deps go into devDependencies is the webpack.dev.dll. Having everything in devDependencies basically negates the current webpack dll setup, and if you point the entry to the devDependencies, it blows up in fantastic fashion and it is trying to package all kinds of things that don't need to be, nor were ever intended to be packed. According to the docs, it is supposed to help the dev environment load faster, I've tried it with and without the dll and my load times are 15 seconds either way (I know it wasn't that slow when I first started this project but it is now). I don't know, just throwing that out there.

aguynamedben:
Re: the sourcemaps, they're I noticed that too when un-asaring my app. It happens with ERB on `master`.

image

I will file a separate issue for that just to track it separately. It's probably the electron-builder config not being strict enough? https://github.com/electron-react-boilerplate/electron-react-boilerplate/blob/master/package.json#L62

For our app, we upload sourcemaps to Sentry in a dev-only script as part of our `yarn release` script. Our sourcemaps are 4MB total so that will be nice to knock off our download size. 👍

aguynamedben:
Re: the Webpack DLL I haven't understood that feature of ERB and am indifferent about the future of that. It looks like the purpose is to build vendor JavaScript so that when HMR happens it doesn't have to reload the vendor JS.

In reality with our app, we're always having to restart the app when we make changes in development because development mostly involves lots of core changes to the data structures (i.e. HMR reloads on a change and then our search index is all of the sudden `undefined` or something, etc.). We haven't spent time tiptoeing around making sure we don't break HRM with application logic.

Issue #2019, opened 2018-11-20
`yarn start` fails with `SyntaxError: Unexpected token import`


gabriel:
`yarn start` fails with `SyntaxError: Unexpected token import`.

## Prerequisites

- [x] Using yarn
- [x] Using an up-to-date master branch
- [x] Using latest version of devtools. See [wiki for howto update](https://github.com/electron-react-boilerplate/electron-react-boilerplate/wiki/DevTools)
- [x] Link to stacktrace in a Gist (for bugs)
- [x] For issue in production release, devtools output of `DEBUG_PROD=true yarn build && yarn start`
- [x] Tried solutions mentioned in [#400](https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/400)

## Expected Behavior

`yarn start` should run the app

## Current Behavior

`yarn start` errors  with `SyntaxError: Unexpected token import`

## Possible Solution

Maybe something to do with babel?

## Steps to Reproduce (for bugs)

1. git checkout https://github.com/electron-react-boilerplate/electron-react-boilerplate
2. yarn install
3. DEBUG_PROD=true yarn build && yarn start

## Context

No changes, clean checkout.

## Your Environment

- Node version : v10.12.0
- Version or Branch used : master
- Operating System and version : macos 10.14.1
- Link to your project : n/a

vikr01:
This was already fixed in #1991. Seems it was merged into [`next`](https://github.com/electron-react-boilerplate/electron-react-boilerplate/tree/next).

PR #2020, opened 2018-11-20
Fix start in package.json

gabriel:
Fixes start in package.json to access app/main.prod.js properly.

Fixes issue #2019.


gabriel:
Fixed in next

PR #2026, opened 2018-11-23
Fix incorrect package position

hirokimoto:
I think **connected-react-router** shouldn't be placed on dev dependencies.


amilajack:
Thanks!

PR #2044, opened 2018-12-04
Add preinstall script for checking if yarn was used

vikr01:


LordZardeck:
Umm, this is causing problems. I clone this, run `yarn install`, and get the error:

```
internal/modules/cjs/loader.js:582
    throw err;
    ^

Error: Cannot find module 'chalk'
```

You included a `require` statement during pre-install for a module that needs installed...lol

Not sure what the solution is other than using postinstall or removing the `chalk` requirement. Can't even run `npm install`

LordZardeck:
@vikr01 I've looked, and I don't see anywhere mentioning that `chalk` comes with node. It's not on their docs as far as I can tell, and a quick google search returned nothing. May I ask where you saw it was included?

vikr01:
@LordZardeck Interesting. I'm not sure where I've seen it, maybe I was wrong. Before merging we tried this with a fresh clone and everything worked fine (maybe due to `yarn`'s caching?). I'll submit a PR to remove the require. Thanks for pointing it out.

LordZardeck:
Ok, glad it was caught before it was in the wild! Not going to lie, you got me pretty excited thinking it was part of the core. That would have been an amazing addition!

PR #2046, opened 2018-12-04
Use multiple issue templates

vikr01:
One for bug report and another for feature request. Left feature request blank for now, but the button should still show up. See [renovatebot/renovate](https://github.com/renovatebot/renovate) for an example.


amilajack:
How about one for 'other'?

PR #2137, opened 2019-01-18
Added window re-activation for MacOS

Foxhoundn:
On MacOS, when you close the window using the "x" close button, you cannot re-open the app unless right clicking on the app icon in dock and selecting "Quit" and then start it again. This change re-creates the window when clicking on the dock icon after the window has been soft-closed.


AmauriAires:
See https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/2122

vikr01:
@AmauriAires Comparing to #2137, this is much more concise. No extra variables are added besides the new function, and the same function is re-used in both `ready` and `activate` (with a check before) states.

PR #2153, opened 2019-02-20
add PLAIN_HMR environment variable option

### LINKED ISSUE #2150: [Feature request] option to turn off react-hot-loader and use simple remounting of root component
I've dealt with so many issues with react-hot-loader that I don't really trust it or think it's necessary.  In most of my projects I just add a `module.hot` hook to reload my `Root` component and remount it with `key={++reloads}`.  Would you be open to a PR for an environment variable to switch to this behavior?

jedwards1211:
Running `yarn dev` with `PLAIN_HMR=1` (any non-empty string) will disable `react-hot-loader` and rely
solely on existing code that remounts the `Root` component with plain HMR.

This is a quick escape hatch to enable someone to keep working if they encounter an issue like bugs in the code output by `react-hot-loader`'s babel plugin.  Ever since I encountered an issue like that in the past, I decided plain HMR was good enough for me.

fix #2150

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import { render } from 'react-dom';
 import { AppContainer as ReactHotAppContainer } from 'react-hot-loader';
 import Root from './containers/Root';
--- a/app/index.js
+++ b/app/index.js
@@ -1,11 +1,15 @@
 import React from 'react';
 import { render } from 'react-dom';
-import { AppContainer } from 'react-hot-loader';
+import { AppContainer as ReactHotAppContainer } from 'react-hot-loader';
 import Root from './containers/Root';
 import { configureStore, history } from './store/configureStore';
 import './app.global.css';
 
 const store = configureStore();
+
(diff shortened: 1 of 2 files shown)


app/index.js
```diff
@@ -1,12 +1,16 @@
 import React from 'react';
```
vikr01:
```suggestion
import React, { Fragment } from 'react';
```

app/index.js
```diff
@@ -1,12 +1,16 @@
...
-import { AppContainer } from 'react-hot-loader';
+import { AppContainer as ReactHotAppContainer } from 'react-hot-loader';
 import Root from './containers/Root';
 import { configureStore, history } from './store/configureStore';
 import './app.global.css';
 
 const store = configureStore();
 
+const AppContainer = process.env.PLAIN_HMR
+  ? React.Fragment
```
vikr01: Would prefer importing `Fragment` explicitly.
```suggestion
  ? Fragment
```

jedwards1211:
Just saw these suggestions. Sounds good, I'll apply them soon.

PR #2159, opened 2019-02-25
fix typo in package.json

Kilian:
boilerplte -> boilerplate


vikr01:
Noticed this a while ago but completely forgot about it, thanks for the PR!

PR #2190, opened 2019-04-12
Loading routes using React.lazy

### LINKED ISSUE #1888: PR: Lazy load routes using React.lazy
https://twitter.com/TheLarkInn/status/1056729742437543936
https://github.com/reactjs/rfcs/blob/gaearon-patch-2/text/0000-lazy.md

fxnoob:
Closes #1888

### DIFF
--- a/app/Routes.js
+++ b/app/Routes.js
@@ -2,8 +2,8 @@ import React, { lazy, Suspense } from 'react';
 import { Switch, Route } from 'react-router';
 import routes from './constants/routes';
 import App from './containers/App';
+import HomePage from './containers/HomePage';
 
-const HomePage = lazy(() => import('./containers/HomePage'));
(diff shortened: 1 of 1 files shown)


amilajack:
I think a nice addition to this would be eagerly loading the HomePage and lazily loading the CounterPage. I'm assuming the current implementation has a flash of content. Is that true?

fxnoob:
Oh, yes. Current implementation has a flash of content.

fxnoob:
Hi @amilajack. It still has a flash of content. How can I resolve this?

amilajack:
I'm not sure because I haven't used these features before.

fxnoob:
Hi @amilajack , 'flash of content' problem only occurring in dev mode as css is bundled with js and that is loaded dynamically. In prod mode 'mini-css-extract-plugin' plugin is use. So there will be no such problem in prod mode.

Issue #2197, opened 2019-04-25
Wrong user data app folder name in dev mode


kevincastejon:
The dev mode application folder has a wrong name ('Electron' instead of the app name), so app.getPath('userData') leads to "[...]/Roaming/Electron" instead of "[...]/Roaming/ElectronReact".
The production mode has the right expected behaviour.

## Expected Behavior

Typing yarn dev create an application folder into the OS 'userData' location which name is the same as the application (as in production mode behaviour).

## Current Behavior

Typing yarn dev will create an application folder into the OS 'userData' location, with a wrong name (Electron instead of the app name).

## Possible Solution

Partial solution : https://stackoverflow.com/questions/35630934/app-getpathuserdata-seems-to-give-the-wrong-path
Add the following package.json file into the 'app' directory :
{
  "name": "electron-react-boilerplate",
  "main": "./main.dev.js"
}
This will generate the application folder with the app name, but an 'Electron' folder is generated as well (itself containing a 'logs' folder)

## Steps to Reproduce (for bugs)

1.
clone the repo
2.
install dependencies (yarn)
3.
launch dev mode (yarn dev)
4.
log Electron.app.getPath('userData') (returns DRIVE_LETTER:\Users\USERNAME\AppData\Roaming\Electron for windows)

## Context

Storing data into an application folder named 'Electron' could cause conflicts when multiple projects are using the same boilerplate, and so the same folder into 'AppData/Roaming'

## Your Environment

- Node version : 10.15.1
- Version or Branch used : master
- Operating System and version : windows 10

clementAC:
This issue cannot be closed and need further investigation on how to solve it. 
As @kevincastejon  said the userData folder is always named Electron on dev environment and it cause conflict when creating several project with ERB. 

I'm not a specialist but what cause the first generation of the pakcage.json to be false?

In a logical point of view it's a bit absurd to regenerate on the flight the package.json without correcting the root cause of the malformation.

PR #2199, opened 2019-04-26
Fix Dev mode uses generic Electron application folder userData

kevincastejon:
Hi friends,
this is my first PR so don't hesitate to edit it

- Added a "preGenerateDevPackage.js" script file which generates a package.json files on the main.dev.js folder, with a modified "productName" field ( productName: appName_dev )
- Added code in main.dev.js that, in dev mode, deletes the generated package.json file once the app is launched, so it doesnt interfere with the linter
- Changed two scripts in the original package.json file:
* "dev": added a call to the script described above
* "start-main-dev-": removed the "main.dev.js" file name so electron command just points to the folder containing the generated package.json file described above and creates a 'userData' application folder with the right name

Fixes issue #2197.


lePioo:
The "delete" phase is hacky and can not be called if the "yarn dev" is aborted by the user...
That would leads the "dev" package.json file to remain on the app folder making crazy our linter...
An alternative approach would be to prevent the linter from READING RULES from this package.json file and let it be always, but I can't be find such an option.
Another approach would be to specify to electron command the app name but I can't find such options either.

PR #2201, opened 2019-04-30
Adds Electron-Unhandled

Slapbox:
This is a simple change so this should just work, but I have not tested it. I just made the edits in-browser.

This should reduce headaches for users of the boilerplate and for maintainers too via a reduction in issues.

In my project I hide the errors in production with `showDialog: false`:

```
unhandled({
    showDialog: false,
    logger: log.error,
  });
```
but to make it noob-friendly and obvious what has gone wrong without needing to consult log files, I've left the default to show errors in production.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,9 +1,18 @@
 import React, { Fragment } from 'react';
 import { render } from 'react-dom';
 import { AppContainer as ReactHotAppContainer } from 'react-hot-loader';
+import log from 'electron-log';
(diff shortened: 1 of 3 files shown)


Slapbox:
@amilajack I didn't realize I didn't follow the `prettier` convention on this PR, but it should be good to go. 

That said, it's failing in Travis CI. I can't understand how or why this change would cause node-gyp problems on those tests. Any thoughts?

Issue #2205, opened 2019-05-10
Cannot call createRootReducer with history bound to history because IHashHistory [1] is incompatible with History [2].


akshayjp123:
`yarn run v1.15.2
$ flow
Launching Flow server for /Users/akshay/Desktop/electron-react-boilerplate
Spawned flow server (pid=63176)
Logs will go to /private/tmp/flow/zSUserszSakshayzSDesktopzSelectron-react-boilerplate.log
Monitor logs will go to /private/tmp/flow/zSUserszSakshayzSDesktopzSelectron-react-boilerplate.monitor_log
Please wait. Server is initializing (waiting for Watchman - giving up in 16 secoError ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ app/store/configureStore.prod.js:10:39

Cannot call createRootReducer with history bound to history because
IHashHistory [1] is incompatible with History [2].

     app/store/configureStore.prod.js
       7│ import type { counterStateType } from '../reducers/types';
       8│
       9│ const history = createHashHistory();
      10│ const rootReducer = createRootReducer(history);
      11│ const router = routerMiddleware(history);
      12│ const enhancer = applyMiddleware(thunk, router);
      13│

     flow-typed/npm/history_v4.9.x.js
 [1] 125│   declare function createHashHistory(opts?: HashHistoryOpts): HashHistory;

     app/reducers/index.js
 [2]   6│ export default function createRootReducer(history: History) {

Found 1 error
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
`

BikramDeka:
How the solve the above bug

saulwiggin:
I'm getting a similar error when I try to configure the store in react redux. I get the flow error above. 

```
const history = createHashHistory();
const rootReducer = createRootReducer(history);
const router = routerMiddleware(history);
const sagaMiddleware = createSagaMiddleware();
const enhancer = applyMiddleware(sagaMiddleware, thunk, router);

function configureStore(initialState) {
  const store = createStore<*, *, *>(rootReducer, initialState, enhancer);
  sagaMiddleware.run(sagas);
  return store;
}

```

saulwiggin:
This is still a live issue. i've tried the History type in createRootReducer

`export default function createRootReducer(history: History) {
[...]

PR #2206, opened 2019-05-11
Added Route level code splitting using React loadable

akshayjp123:
I am lazy loading only the counter page because its of no use lazy loading home page if that is the first page that loads up when the application opens up.


akshayjp123:
Can anybody tell me why my build is failing with Travis? One with the windows OS fails saying no internet connection and other two with mac and linux have problem with make.

akshayjp123:
Just by adding the IHashHistory as the type in createRootReducer function and it will solve the problem referenced to #2205

PR #2215, opened 2019-06-06
Fix fonts not packaged in prod when using SASS

julienma:
Based on https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/1426#issuecomment-458824975.

I had the same issue (FontAwesome 5's fonts not being packaged in prod when importing in my `app.global.scss` => icons displayed an empty square), and this fixed it.
I'm not at ease with webpack, so not 100% sure if this is a good or bad idea.


entozoon:
I suggest adding the publicPath option also to the block below it, the one that goes `compile all other .scss files` ...
So that it can import font files in sass files other than the *.global.scss ones

julienma:
> So that it can import font files in sass files other than the *.global.scss ones

I can't really explain why, but it already works fine.
My fonts are loaded in a `fonts.scss` file, itself required from an `app.global.scss`.

Is that the use-case you had in mind?

entozoon:
Almost, my component imports a `fonts.scss` file. So no global file involved, and it ends up using that second block of test code in the webpack config.

julienma:
Ha, gotcha.
Well, I can certainly add that, but not sure how it'll affect everything else.

WDYT @amilajack @vikr01?

julienma:
Bump?

entozoon:
Bumpingtons

amilajack:
This change breaks some apps and makes css work for others. I'll need to find the cases where this breaks certain apps and then I'll back to you

julienma:
Hey @amilajack, has this been fixed? Or did you find any way to make this (package fonts in prod when using sass) work with the current config?
Thanks!

PR #2220, opened 2019-06-11
Dependency upgrades

jooohhn:
- Upgraded semver patch-compliant dependencies
- Upgraded to react-redux@7.1.0 to support React Hooks
- Upgraded dev configs to use React-🔥-Dom. Allows use of React Hooks during RHL and is more resilient to js errors

### DIFF
--- a/app/containers/Root.js
+++ b/app/containers/Root.js
@@ -1,7 +1,8 @@
 // @flow
-import React, { Component } from 'react';
+import React from 'react';
 import { Provider } from 'react-redux';
 import { ConnectedRouter } from 'connected-react-router';
+import { hot } from 'react-hot-loader/root';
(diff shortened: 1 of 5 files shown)


julienma:
I did not test it myself, but [connected-react-router](https://github.com/supasate/connected-react-router) states that:
> v6.0.0 requires React v16.4.0 and React Redux v6.0. (**It does not support React Redux v7.0 yet**. We're looking into the compability issue.)

jooohhn:
@julienma Good find. The connected-react-router team recently [updated](https://github.com/supasate/connected-react-router/commit/2c2feaa51a1dc399042d1f651519f0a1f2bc6fc6) so that it should support react-redux@7.0 now

amilajack:
Thanks @jooohhn!

PR #2224, opened 2019-07-20
docs: update branch to clone from

nodox:
The instructions say to pull from `master` but it seems that `next` is the default advertised branch. Have we thought about updating the docs or is this intentional?


amilajack:
This was intentional

PR #2228, opened 2019-08-06
Update package.json

pedrocarlos-ti:
Updating node-sass in package.json to use version ^4.12.0 for Nodejs latest version > 12.


amilajack:
Need to update lockfile as well

Issue #2234, opened 2019-08-17
Syntax Error on Package


ivanjeremic:
```SyntaxError: Unexpected token ﻿ in JSON at position 0
SyntaxError: Unexpected token ﻿ in JSON at position 0
    at JSON.parse (<anonymous>)
    at then.it (D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\app-builder-lib\src\util\packageDependencies.ts:55:22)
    at tryCatcher (D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\bluebird\js\release\promise.js:517:31)
    at Promise._settlePromise (D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\bluebird\js\release\promise.js:574:18)
    at Promise._settlePromise0 (D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\bluebird\js\release\promise.js:619:10)
    at Promise._settlePromises (D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\bluebird\js\release\promise.js:699:18)
    at Promise._fulfill (D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\bluebird\js\release\promise.js:643:18)
    at D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\bluebird\js\release\nodeback.js:42:21
    at D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\graceful-fs\graceful-fs.js:90:16
    at FSReqWrap.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:53:3)
From previous event:
    at readJson (D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\app-builder-lib\src\util\packageDependencies.ts:55:6)
    at D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\app-builder-lib\src\util\packageDependencies.ts:224:61
    at Generator.next (<anonymous>)
    at callback (D:\Development\Apps\ElectronOnlyApps\mausys\node_modules\graceful-fs\polyfills.js:295:20)
    at FSReqWrap.oncomplete (fs.js:154:5)
From previous event:
[...]

PR #2267, opened 2019-08-20
Update electron-builder

kas:
Solves #2234.


amilajack:
Do you know why CI is failing?

kas:
I don’t have access to a Linux machine either at the moment, but it seems like an issue others have been encountering recently: https://stackoverflow.com/questions/55674746/travis-sh-0-cant-open-etc-init-d-xvfb

I’m not very familiar with Travis CI but it seems like an easy fix.

julienma:
Issue and fix still valid.

PR #2232, opened 2019-08-14
add electron-redux support with simple demo

msansoni:
An operational and functional implementation of electron-redux in the boilerplate. With a simple (albeit pointless) demonstration increasing and decreasing the counter from the main process.

Following a lot of the discussion and work from #1920 & #1921,@amilajack, @NclsL, @Slapbox and following some of the implementation of electron-redux in Timesheets by electron-redux author @hardchor.

Please feel free to critique, would appreciate anyone else stress testing my solution!

I forked from next -> rather than master, @amilajack let me know if you'd rather I forked from master, or merged back into next rather than this branch.

PR #2281, opened 2019-10-10
add electron-redux support with simple demo

msansoni:
An operational and functional implementation of electron-redux in the boilerplate. With a simple (albeit pointless) demonstration increasing and decreasing the counter from the main process.

Following a lot of the discussion and work from #1920 & #1921,@amilajack, @NclsL, @Slapbox and following some of the implementation of electron-redux in Timesheets by electron-redux author @hardchor.

Please feel free to critique, would appreciate anyone else stress testing my solution!

@amilajack I re-forked from /next rather than rebase, as it was simpler than merging conflicts on the next branch - as a result #2232 can be closed, as this supercedes the request.

I've tried updating the tests as well to reflect the electron-redux integration, however I need some assistance reviewing how to define the spec, with the actions now being exectured from the main process due to the electron-redux actions lifecycle.


msansoni:
An extension on the PR above, using @jp887's implementation of a second browser window, the commit on [multiple-renderers branch](https://github.com/msansoni/electron-react-boilerplate/commit/963955acae7a32e8afd36448bae49004ecd16b87) demonstrates the benefit of electron-redux for a common store across multiple browser windows.

amilajack:
Thanks for the PR! Would love if you could rebase this against master so it is easier to review.

amilajack:
Thanks for the PR! Would love if you could rebase this against master so it is easier to review.

Slapbox:
Just an FYI, `electron-redux` can make debugging really difficult in some circumstances because it can invoke function calls as a result of `IPC`, which makes a mess of call stacks, since you might have a hard time tracing down why something was called. Not a common issue with good design, but a possible support load concern for the maintainers of the library. At the same time, I don't see any easy way to make this an opt-in/opt-out feature.

It also seems to occasionally freeze developer tools if you try to inspect a function invoked by an `IPC` channel, although this is a more general issue I'm facing with Electron 6 that might be unrelated, or might just be more likely to occur with `electron-redux`. I didn't see this particular issue in Electron 4.

amilajack:
@Slapbox thanks for pointing this out! In that case I'll close this PR. I think we should also report those issues to the `electron-redux` repo so they're aware of them.

Slapbox:
I don't know that the second issue is caused or even exacerbated by `electron-redux`, but it does seem like a possible reason. I've filed an issue on the Electron tracker about this, but no one else seems to have the issue, so I've been trying too rationalize out what it might be based on what we're doing that's a bit uncommon. All I came up with was `electron-redux` and using a `preload` script.

The worsening of debugging call stacks is definitely related to the fact that `ipc` calls interrupt and obfuscate what would be a nice clean call stack if it was all done locally. Because `electron-redux` uses _synchronous_ `IPC` calls, debugging in general can be made more difficult, as interrupting a synchronous `IPC` request can cause unusual behavior like freezing or crashing (not always.) I think `electron-redux` is great and we love it despite the several problems. My fear is mostly for your sanity being on the receiving end of support requests.

Slapbox:
Just thought I'd provide an update since I've just run into a case where I'm having difficulty debugging. Here's a situation where I passed an invalid prop to `tippyjs`  and the call stack is useless:

image

`forwardToMain` and `replayActionRenderer` are two functions used by `electron-redux` to keep the store's state in sync.

msansoni:
Thanks @Slapbox, I have to admit I've come into a few cases where electron-redux struggles with keeping the stores in sync - I resorted to blacklisting most the actions to occur only in the main process then just making it a case of using the built in IPC channels of electron-redux to keep the renderer stores in sync. I've just started using @amilajack's neon bindings and demos for rust, which may make a number of the time consuming processes I ran in seperate processes quicker and reduce the requirement for something like electron-redux. - it may be the best recommendation if people come with similar electron-redux integration issues/queries in the future.

Issue #1767, opened 2018-09-08
New route doesn't work (Only "/" and "/counter")


mouhsnimohamed:
I added a new route to my Route.js but cannot access to it! did I miss something?

here is my code in Route.js  
`<Switch>
      <Route path="/counter" component={CounterPage} />
      <Route path={routes.HOME} component={HomePage} />
      <Route path="/category" component={CategoryPage} />
    </Switch>
`
and Menu.js
`<Link` to="/category">{category.title}</Link>`

zhenghaiyang:
i had the same problem

ZymantasPonelis:
` <Route exact path={routes.HOME} component={Home} />`

adding exact solved the problem for me

talha131:
@ZymantasPonelis thank you. Can you explain why does this issue happen? Is it because of react-router or this boilerplate?

aykutyaman:
From the stack overflow: ```The exact param disables the partial matching for a route and makes sure that it only returns the route if the path is an EXACT match to the current url.``` 
You could look at the [answer](https://stackoverflow.com/a/49162423/1998458) for further explanation

PR #2298, opened 2019-10-25
Add comment so folk know better how to add things

jessedp:
A nudge to help make sure folk don't become frustrated trying to add a new page. Issue #1767 brought me to this.
As a noob, It took me a while to find/realize this. Even better would be some documentation additions or a quick tutorial on how to add a "page".


amilajack:
I think it would be better to add better docs for this

PR #2303, opened 2019-10-30
bump electron version to 7.0.0

fudanglp:


app/main.dev.js
```diff
@@ -70,7 +70,10 @@ app.on('ready', async () => {
   mainWindow = new BrowserWindow({
     show: false,
     width: 1024,
-    height: 728
+    height: 728,
+    webPreferences: {
+      nodeIntegration: true
+    }
```
AElmoznino: Will this "works", it's strongly discouraged by the electron docs, read more here: https://electronjs.org/docs/tutorial/security

AElmoznino:
I like the fact that you're trying to bump it, but there are some [security concerns](https://electronjs.org/docs/tutorial/security) by setting `nodeIntegration: true`. Would be better to have strong security by default IMO.
I tried locally to bump it but couldn't get it to work. Anyone got a working solution to share? 😄

fudanglp:
@AElmoznino I see. I'll look into it.

Issue #2312, opened 2019-11-18
Switch Redux logic setup to use Redux Toolkit


markerikson:
Hi, I'm a Redux maintainer. I'd like to offer a couple suggestions.

### Problem

The current Redux example implementation shows use of reducers with "hand-written" immutable update logic, and has a somewhat complex store configuration. In addition, the project currently uses a "folder-by-type" folder structure.

### Proposed Solution

The project should switch to using the official [Redux Toolkit package](https://redux-toolkit.js.org) to implement the Redux logic.

The project should also consider switching to a "feature-folder" or "ducks" folder structure, rather than a "folder-by-type" structure.

### Additional Details

Redux Toolkit is our new official, opinionated, batteries-included toolset for efficient Redux development.  It includes utilities to simplify several common Redux use cases, including store setup, defining reducers, immutable update logic, and even creating entire "slices" of state at once without writing any action creators or action types by hand.

We are recommending RTK as the default standard way to write Redux logic.  Partly related to that, we are also recommending that developers should prefer to use "feature folders" or "ducks" rather than "folder-by-type".  (Reference: the work-in-progress [Redux "Style Guide" docs page](https://deploy-preview-3626--redux-docs.netlify.com/style-guide/style-guide).)

Since many developers will be starting their projects using `electron-react-boilerplate`, it would be beneficial if Redux Toolkit was included as part of the default project setup, as it will greatly simplify the application code they write.

johncantrell97:
I'll take a stab at this.  Do we want to also switch to the "feature-folder" structure as part of it or do that separately?

reZach:
@johncantrell97 I'd imagine that would be apart of this

gregorychen3:
I'm greatly looking forward to trying out this feature. Any idea when it will be officially released? Any help needed doing so?

PR #2351, opened 2019-12-22
Switch Redux setup to use Redux toolkit (#2312)

johncantrell97:
Finished a pass on updating all the redux code to use the new Redux Toolkit best practices as mentioned by @markerikson here:

https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2312

- Updated store creation to use new configureStore() function.
- Replaced manual actions, action creators, and reducers creation with new createSlice() function.
- Removed redux-thunk dependency as it's included by default in the new toolkit.
- Moved towards a more "feature folder" structure by moving all Counter related code into features/counter instead of spread out in actions, reducers, components folders.

There were a few decisions I made where I wasn't 100% sure if they were the right way to go so please review and make suggestions as you see fit.


app/features/counter/CounterPage.js
```diff
@@ -10,7 +10,7 @@ function mapStateToProps(state) {
 }
 
 function mapDispatchToProps(dispatch) {
```
markerikson: Change this to use [the "object shorthand" form of `mapDispatch`](https://react-redux.js.org/using-react-redux/connect-mapdispatch#defining-mapdispatchtoprops-as-an-object).  In fact, given that `counterActions` is already an object, just pass that directly to `connect`:

```js
export default connect(mapStateToProps, counterActions)(Counter)
```

app/features/counter/counterSlice.js
```diff
@@ -0,0 +1,41 @@
+// @flow
+import { createSlice } from '@reduxjs/toolkit';
+import type { GetState, Dispatch } from '../../reducers/types';
```
markerikson: With TS, we recommend inferring the type of the root reducer as a reusable `RootState` type, and using that with thunks:

https://redux.js.org/recipes/usage-with-typescript#type-checking-reducers

Can that be done with Flow?

app/store/configureStore.dev.js
```diff
@@ -1,23 +1,18 @@
-import { createStore, applyMiddleware, compose } from 'redux';
```
markerikson: Is there really any benefit in having separate files for configuring the store in dev vs prod, especially now that RTK is handling setting up the DevTools Extension?

app/features/counter/CounterPage.js
johncantrell97: makes sense, done.

app/features/counter/counterSlice.js
johncantrell97: Yeah this was the main thing I wasn't really sure about as I went through the redux toolkit guide.  I have never used Flow before (only Typescript) so I'm not really sure about this.

app/store/configureStore.dev.js
johncantrell97: I'm not sure why the original decision was made to separate it into separate files.  I have updated this PR to include a consolidation to a single store.js setup file.  It does still include a couple process.env.NODE_ENV checks though which is what they might have originally been trying to avoid?

johncantrell97:
- Updated PR by consolidating the store setup to a single file used by all environments. 
- Also continued to further the 'feature folder' structure by moving 'Home' to its own feature folder enabling the removal of the components directory completely.

Also, as I was working on this I think I noticed the HMR code for the rootReducer has likely been broken:

The old code here:

() => store.replaceReducer(require('../reducers').default)

expects the ../reducers/index.js module to export default the root reducer itself.  In our case we are exporting a function that creates the rootReducer using a passed in history object.

Should I fix this as part of this PR since I'm already updating this file?

GerardKetuma:
I hope this is merged soon. Wanted to do same, good thing I checked opened PRs first.

amilajack:
I'm on vacation now and I'll be back in about a week. Feel free to ping me in a week's time in case I don't get back to this.

app/features/counter/Counter.js
```diff
@@ -36,23 +36,23 @@ export default class Counter extends Component<Props> {
         <div className={styles.btnGroup}>
           <button
             className={styles.btn}
-            onClick={increment}
+            onClick={() => increment()}
```
reZach: Probably not a big deal, but calling functions in his way create new references every time the component re-renders.

amilajack: I agree with. @johncantrell97 Is this a necessary change?

test/actions/counter.spec.js
```diff
@@ -1,43 +0,0 @@
-import { spy } from 'sinon';
```
amilajack: What was this test removed?

app/features/counter/Counter.js
markerikson: It is actually rather necessary.

By default, the action creators generated by Redux Toolkit accept a single argument, which becomes `action.payload`.  `onClick` passes a React event object to the click handler.  So, `onClick={increment}` will end up passing the event object as the action payload.

This becomes an issue because RTK also adds a "serializability check" middleware to the store setup by default, warning about any non-serializable values in actions and state.  The event objects aren't serializable, so RTK will warn whenever this is clicked.

amilajack: Thanks for the clear explanation. Makes a lot of sense.

reZach: Thanks for explaining this @markerikson.

ivanjeremic:
How is it going? Any plans on merging this soon? Is there anything I can help with?

steveetm:
I really like this react-toolkit idea but as the core electron-react-boilerplate moved to use ts and react-hooks don't you guys want to use `useDispatch` and `useSelector` instead `connect` instead of this wrapping component approach? Personally I found it a bit harder to grasp all the concept, `useSelector` a lot cleaner(for me). I think I can send you a PR in the next days if you are willing to review:)

markerikson:
I'd be on board with using hooks as the default approach in the boilerplate.  We're doing that in our Create-React-App templates, and I'll be teaching hooks as the default approach in the new "Quick Start" tutorial page I'm about to write.

istrapi:
Yes I'm 100% for hooks only approach in this boilerplate.

amilajack:
@ivanjeremic @steveetm This was not merged because of no response from the PR author. I'd love to review a PR that makes these changes

PR #2361, opened 2020-01-06
Clarify publicPath setting

eric-burel:
Seems like a trivial change but the "/" at the end of the path got me wrong when updating an old ERB app to Webpack 4, and debug tools do not catch this correctly (wrong `url` in css do not appear in the network tab or in the console so it's hard to detect).

By the way the `module` setup of the CSS seems to somehow break font awesome 4 import. I know this is probably a wontfix since you updated to font awesome v5. I still don't understand why it doesn't work but that's not a big deal.

### DIFF
--- a/configs/webpack.config.renderer.dev.babel.js
+++ b/configs/webpack.config.renderer.dev.babel.js
@@ -57,7 +57,7 @@ export default merge.smart(baseConfig, {
   ],
 
   output: {
-    publicPath: `http://localhost:${port}/dist/`,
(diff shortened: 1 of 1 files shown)


eric-burel:
Stale bot is by far my least favorite bot

eric-burel:
Not stale

amilajack:
Can you please clarify this change

eric-burel:
As far as I remember, it prevents user from forgetting the "\", which is a mistake I made and was not obvious to debug.

PR #2382, opened 2020-02-20
added sqlite config and redux middleware

gterre:


amilajack:
You probably meant to PR this to another rerpo

PR #2401, opened 2020-03-06
Update package.json mac build option

PathToLife:
Hi, I'm proposing adding macOS `package-mac` build option into the package.json

This follows the pre-existing convention where they have specified individual builds for windows / linux.

Personally, I spent ~8mins trying to work out how to build to mac only.
Ended up running `package-linux`, `package-all`, `package`, downloaded 100+ mb of build stuff, then installed `brew install rpm` to get `package-linux` running.

I was wondering what might have prevented a new person from needlessly doing all that when experimenting.

I popped over to https://www.electron.build/configuration/configuration#configuration.

Decided that adding `package-mac` might have been better, so here I am :)


amilajack:
Thanks for this!

PR #2416, opened 2020-03-22
Fix TypeScript errors in menu.ts

rqbik:
Apparently, MenuItemConstructorOptions.selector is a Darwin-only attribute.

image

TypeScript detects it and displays an error. To fix this error we can cast every menu that is using it to MenuItemConstructorOptions.

### DIFF
--- a/app/menu.ts
+++ b/app/menu.ts
@@ -122,7 +122,7 @@ export default class MenuBuilder {
         }
       ]
     };
-    const subMenuViewProd = {
+    const subMenuViewProd: MenuItemConstructorOptions = {
       label: 'View',
       submenu: [
         {
--- a/app/menu.ts
+++ b/app/menu.ts
@@ -147,7 +147,7 @@ export default class MenuBuilder {
         { label: 'Bring All to Front', selector: 'arrangeInFront:' }
       ]
     } as MenuItemConstructorOptions;
-    const subMenuHelp = {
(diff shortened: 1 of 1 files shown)


app/menu.ts
```diff
@@ -80,11 +79,10 @@ export default class MenuBuilder {
           }
         }
       ]
-    };
-    const subMenuEdit: MenuItemConstructorOptions = {
+    } as MenuItemConstructorOptions;
```
steveetm: I don't really like this, this will treat every kind of menu definition as a proper one. E.g.
```
const faultSubMenuAbout = {
labelWithType: 'Electron';
suubMenu: [],
} as MenuItemConstructorOptions;
```
No ts warnings:(

How about declaring a darwin specific type?
```
interface DarwinMenuItemConstructorOptions extends MenuItemConstructorOptions{
  selector?: string;
  submenu?: (DarwinMenuItemConstructorOptions[]) | (Menu);
}

const faultySubMenuAbout: DarwinMenuItemConstructorOptions {
typeInLable: 'Electron';
}
```

amilajack: I like this solution a lot more!

rqbik:
Yeah, I didn't think about that... Fixed it according to your request.

amilajack:
Thanks for the fix!

PR #2443, opened 2020-04-24
Remove extraneous webpack-dev-server client from entries.

steveetm:
As `webpack-dev-server` used for `yarn dev`, adding webpack-dev-server/client is not needed, only if we instantiate webpack-dev-server via nodejs api, afaik.


steveetm:
Just realised HMR is not working for reducers and hot/only-dev-server is extraneous too, I think will review HMR and resend this.

PR #2469, opened 2020-05-21
Updated electron 9.0.0

null-none:


johnstew:
This should get merged in otherwise you get errors that are only solvable by upgrading on your own.

johnstew:
On second thought, I'm not sure we want to delete the yarn.lock

null-none:
@johnstew about `yarn.lock` ok

DmitryMK:
This update does not allow to use react/redux devtools.

PR #2471, opened 2020-05-27
Update electron-devtools-installer to v3

asmeikal:
Builds for windows made with version 2 of `electron-devtools-installer` did not start if `DEBUG_PROD` was set to `true`.

See [this issue](https://github.com/MarshallOfSound/electron-devtools-installer/issues/72) and similar issues on the `electron-devtools-installer` repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -217,7 +217,7 @@
     "detect-port": "^1.3.0",
     "electron": "7.1.13",
     "electron-builder": "^22.3.6",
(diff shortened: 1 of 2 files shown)


amilajack:
Thanks for the PR! Changes like this are usually made to the `next` branch so that I can test them out before releasing a new version. After rebasing against `next`, I can merge this.

PR #2479, opened 2020-06-04
Remove unnecessary yarn ts script

jooohhn:
Reduce verbosity since calling `tsc` directly works just as `yarn ts` would


kyranjamie:
Less is more 👍

PR #2480, opened 2020-06-06
Remove unnecessary file extensions from ts exclude

jooohhn:
Removed `.css` and `.map` files from `tsconfig.json`. They aren't considered anyways during `tsc` execution (according to [typescript documentation](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#details))

> If the "files" and "include" are both left unspecified, the compiler defaults to including all TypeScript (.ts, .d.ts and .tsx) files in the containing directory and subdirectories except those excluded using the "exclude" property. JS files (.js and .jsx) are also included if allowJs is set to true.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -26,14 +26,9 @@
     "test",
     "release",
     "app/main.prod.js",
-    "app/main.prod.js.map",
     "app/renderer.prod.js",
-    "app/renderer.prod.js.map",
-    "app/style.css",
-    "app/style.css.map",
     "app/dist",
     "dll",
-    "app/main.js",
-    "app/main.js.map"
+    "app/main.js"
   ]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -28,7 +28,6 @@
     "app/main.prod.js",
     "app/renderer.prod.js",
     "app/dist",
-    "dll",
-    "app/main.js"
+    "dll"
   ]
 }


tsconfig.json
```diff
@@ -26,14 +26,9 @@
     "test",
     "release",
     "app/main.prod.js",
-    "app/main.prod.js.map",
```
amilajack, line 29: I think `app/main.prod.js.map` is used

tsconfig.json
```diff
@@ -26,14 +26,9 @@
...
     "app/renderer.prod.js",
-    "app/renderer.prod.js.map",
-    "app/style.css",
-    "app/style.css.map",
     "app/dist",
     "dll",
-    "app/main.js",
-    "app/main.js.map"
+    "app/main.js"
```
amilajack, line 32: Does this file exist for you? I can't find it in my local build

jooohhn: nah can't find it, should it be removed too?

tsconfig.json
jooohhn: don't believe so, typescript doesn't care about files other than .ts, .d.ts, .tsx, .js, .jsx(https://github.com/microsoft/TypeScript/issues/10939)

The # of files linted is the same with the old and new tsconfig
image

amilajack: Ahh I see, good point. Thanks for clarifying!

PR #2501, opened 2020-06-28
Updated man.dev file from require to import

sdc224:
Updated `main.dev.ts` file, replaced `require` statements with `import`

Feel free to add something if I miss.


amilajack:
Thanks for the PR!

PR #2503, opened 2020-06-29
Fix for preload entrypoint

### LINKED ISSUE #2395: How to properly set up material-ui with electron-react-boilerplate
See this question: https://stackoverflow.com/questions/60473495/how-to-properly-set-up-material-ui-with-electron-react-boilerplate

murrayju:
Fixes #2395.

I believe that this is a proper fix, and not just a workaround. This probably requires validation from someone more knowledgeable.


peterrogov:
@murrayju thank you very much for your fix! it helped me in my app as well even though I don't use redux and react-navigation. worked like a charm! will test more but so far so good.

PR #2516, opened 2020-07-07
Fix e2e test error

jeean:
**1. App is not showing up**
When I try `yarn build-e2e` and `yarn test-e2e`, app doesn't show up because `app.on('ready')` event is not fired after electron is executed by testcafe. After I use `whenReady()` instead of `on('read')` it works fine.

**2. Webpack Dynamic Imports not working properly**
After electron is launched by testcafe, app can't load `CounterPage` component because script is trying to reference `/app` path instead of `/app/dist/` and showing javascript error below.
```
JavaScript error details:
      Error: ENOENT: no such file or directory, open '/User-Directory-Path/electron-react-boilerplate/app/0.renderer.prod.js'
```
However after I build app with `electron-renderer` target not `electron-preload`, dynamic imports works well.

If you know the better way to fix error about e2e test, please edit the code.

### DIFF
--- a/app/main.dev.ts
+++ b/app/main.dev.ts
@@ -110,7 +110,12 @@ app.on('window-all-closed', () => {
   }
 });
 
-app.on('ready', createWindow);
+if (process.env.E2E_BUILD === 'true') {
+  // eslint-disable-next-line promise/catch-or-return
+  app.whenReady().then(createWindow);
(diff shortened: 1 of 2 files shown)


amilajack:
Also can you change the pr to be against the 'next' branch

Issue #2517, opened 2020-07-07
Redux DevTools not working, showing "No store found. Make sure to follow the instructions."


dabo248:
## Prerequisites

- [X] Using yarn
- [X] Using node 10.x
- [X] Using an up-to-date [`master` branch](https://github.com/electron-react-boilerplate/electron-react-boilerplate/tree/master)
- [X] Using latest version of devtools. See [wiki for howto update](https://electron-react-boilerplate.js.org/docs/dev-tools/)
- [X] Tried solutions mentioned in [#400](https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/400)

## Expected Behavior

Expected Redux DevTools to be usable out of the box.

## Current Behavior

When I select the Redux tab in the dev console, I see one sentence stating "No store found. Make sure to follow the instructions."

## Possible Solution

This issue might be related to these:

https://github.com/electron/electron/issues/23662
https://github.com/electron/electron/issues/24011

## Steps to Reproduce (for bugs)

1. Clone and install project like instructed here: https://github.com/electron-react-boilerplate/electron-react-boilerplate#install

2. Run `yarn dev`

3. Click on the Redux tab in the dev console

4. You should see a grey background with the message: "No store found. Make sure to follow the instructions."

## Context

I'm trying to track and understand my application state with Redux DevTools. 

Redux DevTools is usable with electron-react-boilerplate version 1.1.0, so I'm not quite sure if the recent switch to Redux-Toolkit or the switch to Electron version 9 broke it. 

## Your Environment

- Node version : 10.17.0 and 12.18.0
- Version or Branch used : 1.2.0 and master
- Operating System and version : macOS Catalina (10.15.5)
- Link to your project : Happens with the plain boilerplate project (https://github.com/electron-react-boilerplate/electron-react-boilerplate)

ChALkeR:
https://github.com/electron-react-boilerplate/electron-react-boilerplate/blob/master/app/main.dev.ts#L73 -- this is probably breaking on 9.x due to electron/electron#24011.

pranshuchittora:
I also tried passing the `devTools: true` flag explicitly to the `configureStore` of the `@reduxjs/toolkit` unfortunately no progress. @ChALkeR you are right I think the problem is with the electron.

nikandlv:
Same issue on a fresh generated project

amilajack:
[...]

PR #2525, opened 2020-07-19
v1.3.0

amilajack:
- Fixes E2E tests (#2516 )
- Fixes preload entrypoint (#2503)

PR #2527, opened 2020-07-22
Downgrade Electron

jooohhn:
Reverts breaking electron changes introduced in #2465 

Fixes
- #2517


package.json
```diff
@@ -218,10 +218,10 @@
     "cross-env": "^7.0.0",
     "css-loader": "^3.6.0",
     "detect-port": "^1.3.0",
-    "electron": "9.1.0",
-    "electron-builder": "^22.7.0",
-    "electron-devtools-installer": "^3.1.1",
-    "electron-rebuild": "^1.11.0",
+    "electron": "7.1.13",
```
amilajack, line 221: Let's try to see if we can get this working with `electron@^8`. 7 has a lot of bugs

package.json
```diff
@@ -218,10 +218,10 @@
...
+    "electron-builder": "^22.3.6",
+    "electron-devtools-installer": "^2.2.4",
```
amilajack, line 223: Why were these deps downgraded?

jooohhn: Reverted back to the exact deps before #2525 breaking changes

amilajack: I see. I dont think deps need to be downgraded since the bug was caused with electron. Ideally we could keep as many deps up to date as possible.

nikandlv:
Can we have this merged?

Issue #2523, opened 2020-07-16
scss compiles error


Mad-hu:
![image](https://user-images.githubusercontent.com/13977755/87665972-644ed380-c79a-11ea-853a-24bbca37a398.png)
recurrent trouble
git clone,and change files 
Counter.css->Counter.scss
Counter.tsx ->
```js
import React from 'react';
import { useSelector, useDispatch } from 'react-redux';
// import { Link } from 'react-router-dom';
import './Counter.scss';
// import routes from '../../constants/routes.json';
import {
  // increment,
  // decrement,
  // incrementIfOdd,
  // incrementAsync,
  selectCount,
} from './counterSlice';

export default function Counter() {
  // const dispatch = useDispatch();
  // const value = useSelector(selectCount);
  return (
    <div>

    </div>
  );
}
```
yarn dev ,and error
- [ 1.19.1 ] Using yarn
- [ 12.14.1 ] Using node 10.x

"name": "css-loader",
"version": "3.6.0",
![image](https://user-images.githubusercontent.com/13977755/87666690-a7f60d00-c79b-11ea-992b-29ca79e7a5d3.png)

fondoger:
Same for me

sepehr09:
i have same problem
i do some search and find that `typings-for-css-modules-loader` is deprecated and not maintain any more. i think that's cause this problem.

any solution?

jooohhn:
Working on a fix at the moment, bug seemed to be introduced by dependency upgrades in v1.2.0

Temporary solution would be to use v1.1.0 (#2409)

jooohhn:
Should be fixed and incorporated by next project release. If you need it ASAP check out changes merged into the `next` branch (#2540)

PR #2533, opened 2020-07-29
fix scss compile error (#2523)

sepehr09:
the `typings-for-css-modules-loader` package is deprecated and not maintain any more. and cause below error when using scss files: 

```bash
Module build failed (from ./node_modules/typings-for-css-modules-loader/lib/index.js):
Error: Cannot find module 'css-loader/locals'
```

issue #2523


jooohhn:
Thanks for the PR! I overlooked this PR and merged a similar fix into `next` that is able to keep css modules typing (#2540)

jooohhn:
Thanks for the PR! I overlooked this PR and merged a similar fix into `next` that is able to keep css modules typing (#2540)

PR #2136, opened 2019-01-18
Web enablement: dev and prod

janakg:
Hi, PR for the "Publishing app to web & electron: Issue 80".

I have created a clean way to integrate web distribution and development. Please review the changes so we can close the long pending request from the community.

Thanks.

Issue #80, opened 2015-11-24
Publishing app to web & electron – preferred approach?


t1mmen:
I'm looking to publish my app both to the web, and as a standalone executable. 

Is anyone doing this, and if so, care to share your preferred approach? If not, is this something the maintainers are interested in as part of the repo? I'll try to lend a helping hand in the coming weeks if so.

chentsulin:
I think we should not include web stuff into this boilerplate, it will increase unnecessary complexity in a desktop-only application. But if someone is interested in how to build them together, I can work on a separate repo to demo it in next couple of weeks.

t1mmen:
@chentsulin Makes sense to keep it separate for complexities sake. I'd be *very* interested in a demo of how to, though!

alexhawkins:
I'd be very interested in this as well. @chentsulin  Let me know if you decide to do this. I'm going to see if I can get it to work on my own but any help from anyone would be much appreciated. Shoot me an email @ alex.hawkins@shiphawk.com

flywithmemsl:
Yo @chentsulin, I'm currently working with your boilerplate, it awesome, thank you. 
I'll need website build feature in next two weeks, so if I can help you somehow with it, please tell me.

acao:
I'm doing this with a modified version of this boilerplate.

All i had to do was handle argv from make-webpack-config, and when 'web' is provided then the webpack options.target = web, otherwise the electron renderer as this boilerplate provides.

Deployment is another matter - I just adapted the package.json scripts to a makefile and deploy the web version to s3 (after running a build-web using --target web as described above).

It's actually not very complicated at all. The only thing I have left to is find some way to allow the renderer process to detect whether its being run in electron or in the web, so that for example 'openDialog' function can choose to open a native dialog or material ui dialog.

The one thing that will get you caught is what build you're running in development depending on whether you are developing for electron or web, and same for deployment. Oh and make sure hot-dev-server knows what it needs to do as well.

acao:
Apologies, I forked this project a few months ago and the webpack config has changed significantly since then. I hope that this still helps folks who are looking to get this working!

flywithmemsl:
https://gist.github.com/flywithmemsl/0d9eba3c7e0ebc3dddc3 — works for me
replaced `commonjs2` with `var` in libraryTarget
removed `webpackTargetElectronRenderer`

acao:
[...]

Issue #2512, opened 2020-07-03
Minimal build template


prescottbreeden:
This is really great work that y'all have built here.  Would love a significantly reduced version that isn't pre-opinionated about style sheets, redux, and routing -- it is surprisingly tricky to customize this repo out of the box.

amilajack:
@prescottbreeden There's a lot of interests in dropping support for redux and this is something we'll be looking at in the near future. I was wondering what your thoughts were on why you consider styling and routing opinionated

manutero:
Same here. The work here is astonishing.

In my case, I wanted to remove React-Redux from the equation. Got it but took me a while (and I'm not 100% sure I've left any unused dependency behind).

Now I'm considering even publishing it as a fork "electron-react-boilerplate-no-redux"  xD

tudiantuan:
+1

jooohhn:
General idea is accepted and is being continued in #2553

Issue #2476, opened 2020-05-31
add docs for using JS version of ERB


amilajack:
Closes https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2475

PR #2546, opened 2020-08-06
Enable web: dev and build

rntjr:
This PR is an update to version 1.3.0 of PR #2136 and issue #80.
Dev and build webpacks for the web are configured for typescript, working perfectly.

Need for someone else to validate.


rntjr:
Lockfile fixed.

jooohhn:
Thanks for the PR! My initial thoughts:

Pros:
- Benefits many users since many electron-based apps also have a web version

Cons: 
- Would add complexity to maintaining ERB
- Would add unnecessary bloat for users that don't need it

IMO the last con is particularly important. There's a lot of interest for putting user configurations over project convention (#2476, #2512)

In the case where this PR isn't merged, I strongly think that the content should still be incorporated into the ERB docs since

rntjr:
In fact, maintenance is very complex, especially on large projects.
I am analyzing to develop an ERP tool having a central server and also private servers of the clients themselves. The main problem that is making me abandon Electron for this project is that, whenever there is a private server, I will have to do a different build for it.
But in applications that have a central server it has great usability.

barrynorman:
Is this still working? I was wondering how apps like slack are building a web version. thanks

Unforgiven-wanda:
Any update for getting this to work on the latest boilerplate version?

Issue #2511, opened 2020-07-01
System tray icon


dugiwarc:
System tray icon shows in development mode, but not after packaging the app.
Any workaround to this?

This is how I add the icon 
```
let tray = new Tray(
      nativeImage.createFromPath(
        path.join(__dirname, 'resources', 'icons', '24x24.png')
      )
    );
```

masonlee:
I'm having the same issue.  What do we need to do to get our main process image assets into the packaged app?

masonlee:
I figured it out with help from the [electron-builder docs](https://www.electron.build/configuration/contents#extraresources) and https://stackoverflow.com/a/46033483/530309

I moved my tray icon into a new folder called `[my_project]/extraResources`.

Then I added a definition for "extraResources" in `[my_project]/package.json`.  For me, that looks like:

```
  "build": {
    "extraResources": [
      "./extraResources/**"
    ],
   ...
```

Then, when referencing the tray icon by name, I use a new getAssetPath() function so that it can be found in dev mode and when packaged.  Here's that:

```
import path from 'path';

const EXTRA_RESOURCES_PATH = app.isPackaged
  ? path.join(process.resourcesPath, 'extraResources')
  : path.join(__dirname, '../../extraResources');  // Your relative path may be different!

const getAssetPath = (resourceFilename: string): string => {
  return path.join(EXTRA_RESOURCES_PATH, resourceFilename);
};
```

Hope this helps others who are new to Electron.

paragbaxi:
Does anyone have a repo example that uses erb with a working tray from "yarn package"?

patrickkreuzenort:
My problem with packaging of an embedded sqlite db with prepared data is now solved, thank you very much!

PR #2567, opened 2020-08-31
Provide default app icon

CZYK:
This adds the default app icon to the boilerplate.

Based on https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2511


amilajack:
nice! 🥳

PR #2574, opened 2020-09-01
Async in tests

rharink:
I've added the `@babel/plugin-transform-runtime` to be able to use asyc/await in jest tests.


amilajack (approved):
Thanks!

PR #2586, opened 2020-09-14
adds cross-env as dev dependency

tinjaw:
I tried to use this boilerplate on Windows and it failed. I know cross-env is usually a global, but if it is not on the system already, it is needed.


tinjaw:
Did a commit --amend. I guess I should have done a fresh commit. Will close and resubmit.

PR #2587, opened 2020-09-14
Missing cross env

tinjaw:
I tried to use this boilerplate on Windows and it failed. I know cross-env is usually a global, but if it is not on the system already, it is needed.


tinjaw:
It looks like the Mac test timedout on a network issue.

`error An unexpected error occurred: "https://registry.yarnpkg.com/core-js/-/core-js-2.6.11.tgz: ESOCKETTIMEDOUT".`

PR #2600, opened 2020-09-27
chore: improve devtools installer usage

aprilandjan:
It seems that the [electron-devtools-installer](https://github.com/MarshallOfSound/electron-devtools-installer) supports install multiple extensions when passing first argument as array, so there's no need that we do that `promise.all` mapping. Just a tiny improvement which makes it slightly clearer.


amilajack (approved):

amilajack:
Thanks for this!

PR #2602, opened 2020-09-27
Added .env support for window icon and debug

Stancobridge:
Configuring App window icons and debug is a little bit tedious, I have to look through the files to do that.
What I did

- Add .env support with dot-env 
- .env file in the root folder


amilajack:
I don't have a windows machine so I'm not able to see what this fixes. Can you please clarify what this fixes?

Stancobridge:
> I don't have a windows machine so I'm not able to see what this fixes. Can you please clarify what this fixes

Sorry, my bad I mean app window title and window icon, not actually Windows as OS but window and app window, I made mistake with the 's'

Normally to configure app icon, you need to go to ./app/main.dev.ts and touch some configuration codes, or pasting your icon in ./resources and rename it to icon.png
currently, I added support for .env file in the root directly,
you can simply go to the .env file and edit APP_ICON to the name of the icon u have in ./resources

I also updated how Browser Debugger works, by the default if you run the app the first time the Browser Debugger is shown, you can't control this if you are in development mode, you can't decide if you want it to show or not but currently if you are in development mode, you can set that too in the .env file, by setting APP_DEBUG to true or false.

rahulpsd18:
I second this, not for the icon one, as it's a one time thing for me but for `.env` file support.

Stancobridge:
Am not good with the github test check, I don't know why the test is failing.
If this .env support is merged I will be glad, it will make development easier and the firm am working with is already using it without issues so far, we are only planning to push more update on the .env support and some other updates like file manager support but we will need them to be supported in the core framework so that it won't be breaking in case of any update in the core framework

amilajack:
The window icon is added by electron-builder. I'm not sure why this change is necessary. See https://www.electron.build/icons.html. Can you please clarify.

PR #2613, opened 2020-10-07
Makes git automatically converting LF to CRLF and vice versa depending on system

madnesspie:
Hi! First of all thanks for your excellent work. Its really saves time. 

I faced with line ending issue with setting from `.gitattributes` - `* text eol=lf`. Point is that when git makes force line endings convertion from CRLF to LF it can break down `.exe` file. It wouldn't work with following output. 
```
Program 'Senso_DK3_connector.exe' failed to run: The specified executable is not a valid application for this OS platform.At line:1 char:1
+ C:\Users\alex\Projects\frontend-desktop\app\infrastructure\drivers\dk ...
At line:1 char:1
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + FullyQualifiedErrorId : NativeCommandFailed
``` 

I suggest makes git handle line endings convertion automatically depending on current system to avoid such issue. 

Thanks.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
-* text eol=lf
+* text=auto
 *.png binary
 *.jpg binary
 *.jpeg binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,7 @@
-* text=auto
-*.png binary
-*.jpg binary
-*.jpeg binary
-*.ico binary
-*.icns binary
+*       text    eol=lf
+*.exe   binary
+*.png   binary
+*.jpg   binary
+*.jpeg  binary
+*.ico   binary
+*.icns  binary


.gitattributes
```diff
@@ -1,4 +1,4 @@
-* text eol=lf
+* text=auto
```
amilajack, line 1: I think we should limit `text=auto` to only `.exe` files and keep `eol=lf` for all other files. Setting to auto could break our prettier config.

amilajack (changes requested):

.gitattributes
madnesspie: Okey, I think in this case just mark *.exe as a binary (alias for `-text -diff`) is the solution. I've already tested it and it's work.

amilajack (approved):

amilajack:
Thanks! 🥳

PR #2628, opened 2020-10-22
Generate all other required icons from single icon.png in resource folder 

arslanmughal99:
With this commit you will able to generate all other icons required in `resources` directory from single `icon.png` using script `yarn gen:icons`


amilajack:
Where is the source for the generated binary files in the `bin` directory?

arslanmughal99:
Check these out 
https://github.com/arslanmughal99/erb-assets-generator
https://github.com/arslanmughal99/erb-icons-gen

may be this is not the perfect fit but my point here is to give an idea , Because it is very painful to generate icons each time there is small change from designers team. So there must b some script that generate all icons automatically from single icon.png

arslanmughal99:
The binary is written in Golang

amilajack:
I see. Thanks for creating a PR for this. Ideally the tool could be written in node so that developers could take tweak the tool if necessary. Do you know of any node equivalents?

arslanmughal99:
Yes i think it is possible with [Jimp](https://www.npmjs.com/package/jimp) but i dont know well Jimp will perform .

amilajack:
I think I'd prefer erb-generate using jimp here since. Closing this in favor of a PR that generates icons from jimp

arslanmughal99:
Check [electron-icon-builder](https://www.npmjs.com/package/electron-icon-builder) . This looks more promising to me only you need to change folder structure a little for resources .

arslanmughal99:
i am working on new package and i will update you today 👍

arslanmughal99:
Checkout this one i have removed binaries, [erb-icons-gen](https://github.com/arslanmughal99/erb-icons-gen)

amilajack:
This package seems to have good community support:
https://github.com/jaretburkett/electron-icon-maker

I would prefer to use this library since it seems. What are your thoughts on this?

arslanmughal99:
Yes if you think this works best then you can continue with `electron-icon-maker` , However this feature is optional and does not required any active community to maintain and using `electron-icon-maker` will make folder structure verbose .

arslanmughal99:
Any updates on this ?

amilajack:
It seems like electron-icon-maker seems like the road forward. Closing in favor of a PR for integrating with it.

PR #2691, opened 2020-12-21
Prototype testcafe support for Electron e2e tests

trungutt:
This PR should be considered like a prototype for Testcafé support (Electron e2e testing).

We should have two modes for e2e testing:

- Development mode (no packaging required): we need to have renderer process started (without the main process), and starting e2e testing by `yarn test:e2e:dev`. The advantage is no need for packaging/installing app.

- Production mode (which requires application installation):  we need to have the final application installed on host machine, and starting e2e testing by `yarn test:e2e:prod`. This approach reflects the real interaction between app and users.

This prototype is for Mac, for Windows or Linux it's similar (by replacing the path to the installed binary)

### DIFF
--- a/.erb/configs/webpack.config.renderer.dev.babel.js
+++ b/.erb/configs/webpack.config.renderer.dev.babel.js
@@ -259,15 +259,5 @@ export default merge(baseConfig, {
       verbose: true,
       disableDotRule: false,
     },
-    before() {
-      console.log('Starting Main Process...');
-        spawn('npm', ['run', 'start:main'], {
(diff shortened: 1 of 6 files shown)


amilajack:
You may need to dedupe the yarn.lock file as a workaround. See https://stackoverflow.com/questions/52399839/duplicate-identifier-librarymanagedattributes

trungutt:
I close this PR because I have a random timeout problem with Testcafé when testing Electron browser. It seems like it's an issue from Testcafé

PR #2747, opened 2021-01-27
v2.3.0

amilajack:


emilbonnek (commented):
Is this project intended to only work in bash? the rm command isn't in DOS so this doesn't work on windows

PR #2773, opened 2021-02-21
feat: replace `rm` in npm script with node command

yuqianma:
As `rm` command in win is differ from bash, this commit replaces `rm` with node command. That makes "yarn package" works again.

Another choice is writing a build script to make it platform-agnostic. But an one line command is enough now, right?


amilajack:
If I remember correctly, `rm` in npm scripts is "polyfilled" with a custom node implementation on windows. If this is the case this change isn't necessary

yuqianma:
I thinks your memory is incorrectly. I cannot find any information about the "polyfilled". And I tested on my machine. The command `rm` in npm script throws errors in both powershell and cmd. Or I missed something? Can you provide the reference?

GorlikItsMe:
@yuqianma  Thanks for your commit, That was thing what I looking for.

PR #2784, opened 2021-03-01
Migrate to Memory Router

ivanjeremic:
A Router that keeps the history of your “URL” in memory (does not read or write to the address bar). Environments like React Native use it and I believe this is the right way to do routing with React in Electron to.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
+import { MemoryRouter, Switch, Route } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


amilajack:
If I remember correctly this caused issues in the past with electron. Were you able to extensively test this change?

ivanjeremic:
> If I remember correctly this caused issues in the past with electron. Were you able to extensively test this change?

I was not sadly able to test it because I stopped working on the app, but doing some research I came across this solution and that it helped people in the past and also solved the issue for me in my project I didn't encounter any issues so far.

amilajack:
replaced by https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/2874

PR #2817, opened 2021-03-30
resolved issue with document.write() script injection

CLantigua2:
This PR resolves the following issue.
When running the boilerplate in development, the following warning can be seen in the console:
```bash
[Violation] Avoid using document.write(). https://developers.google.com/web/updates/2016/08/removing-document-write
```
This is due to the usage of document.write() within the index.html file. Per the above doc, document.write() is parser blocking which means even if defer is passed in, the write will still execute and block the page rendering. 

After my fix, I tested that I am still getting the appropriate script tags included with the defer tags as previously implemented without warnings.
image

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -33,11 +33,13 @@
       }
 
       if (scripts.length) {
-        document.write(
-          scripts
-            .map((script) => `<script defer src="${script}"><\/script>`)
-            .join('')
-        );
+        const body = document.querySelector('body');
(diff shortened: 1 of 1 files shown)


davipk:
Just a heads up, this change will only work for people that are using ``did-finish-load`` listener to show the app when it's done rendering. For some reason, the main seems to show the renderer even when it's not been fully rendered yet if you apply this change while listening to ``ready-to-show``.

Issue #2794, opened 2021-03-14
Not working with electron 12


TimvdEijnden:
When upgrading to electron 12 the index.html does not have access to `require` anymore and the application is not started.
An error is shown instead

## Prerequisites

- [x] Using yarn
- [x] Using an up-to-date [`master` branch](https://github.com/electron-react-boilerplate/electron-react-boilerplate/tree/master)
- [x] Using latest version of devtools. [Check the docs for how to update](https://electron-react-boilerplate.js.org/docs/dev-tools/)
- [x] Tried solutions mentioned in [#400](https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/400)
- [ ] For issue in production release, add devtools output of `DEBUG_PROD=true yarn build && yarn start`

## Expected Behavior

electron-react-boilerplate page rendered 

## Current Behavior

```
Uncaught ReferenceError: process is not defined
    at index.html:8
    at index.html:15
```

## Steps to Reproduce
1. git clone https://github.com/electron-react-boilerplate/electron-react-boilerplate.git
2. yarn 
3. yarn add electron@12.0.0
4.  yarn start

## Possible Solution (Not obligatory)

Downgrade to latest electron 11.3.0

## Your Environment

- Node version : v14.15.4
- electron-react-boilerplate version or branch : master 
- Operating System and version : macos 11.2.2
- Link to your project : n/a

TimvdEijnden:
Solved by adding 

`contextIsolation: false,` 

to the webPreferences

Issue #2654, opened 2020-11-12
Add support for contextIsolation


githubBrandon:
`contextIsolation` will default to `true` in Electron 12, but this repo doesn't currently support it. (The page doesn't render correctly when setting `contextIsolation: true`.)

It would also help to secure apps that are based on this repo. I see that it was brought up before in #1309, but now that `contextIsolation` is going to be the default, it seems appropriate to revisit the possibility.

amilajack:
@githubBrandon thanks for bringing this up! I'll put this on the roadmap for the next upcoming release!

reZach:
To note, @githubBrandon (also @amilajack) - adding support for contextIsolation requires that any package your application uses and supports contextIsolation. When I was working on electron apps, they _were not_. 

I ended up building [`secure-electron-template`](https://github.com/reZach/secure-electron-template) to address this, so you can look to my repository for inspiration if you'd like.

amilajack:
Fixed in v3

PR #2825, opened 2021-04-03
Updated to Electron 12 with contextIsolation enforcement

davipk:
This PR updates the boilerplate to electron 12 as well as enforcing ``contextIsolation`` that has been set to ``true`` as default due to security reasons listed [here](https://github.com/electron/electron/blob/master/docs/tutorial/security.md#3-enable-context-isolation-for-remote-content).

This should resolve issues #2794, #2766, and #2654.

Please let me know if there's any changes here that might be improved and I would be happy to make the change as requested.

Thanks!

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -26,7 +26,7 @@
         scripts.push('../.erb/dll/renderer.dev.dll.js');
 
         // Dynamically insert the bundled app script in the renderer process
-        const port = dev.port ?? 1212;
+        const port = dev.port || 1212;
         scripts.push(`http://localhost:${port}/dist/renderer.dev.js`);
       } else {
         scripts.push('./dist/renderer.prod.js');
--- a/src/index.html
+++ b/src/index.html
@@ -5,7 +5,7 @@
     <title>Hello Electron React!</title>
     <script>
       (() => {
-        if (process?.env?.NODE_ENV !== 'development') {
+        if (!dev.isRunning) {
           const link = document.createElement('link');
           link.rel = 'stylesheet';
           link.href = './dist/style.css';
@@ -20,13 +20,13 @@
     <script>
       const scripts = [];
 
-      if (process.env.NODE_ENV === 'development') {
+      if (dev.isRunning) {
         // Dynamically insert the DLL script in development env in the
         // renderer process
         scripts.push('../.erb/dll/renderer.dev.dll.js');
 
         // Dynamically insert the bundled app script in the renderer process
-        const port = process.env.PORT || 1212;
+        const port = dev.port ?? 1212;
         scripts.push(`http://localhost:${port}/dist/renderer.dev.js`);
       } else {
         scripts.push('./dist/renderer.prod.js');
--- a/.erb/configs/webpack.config.base.js
+++ b/.erb/configs/webpack.config.base.js
@@ -26,8 +26,7 @@ export default {
 
   output: {
     path: path.join(__dirname, '../../src'),
-    // https://github.com/webpack/webpack/issues/1114
-    libraryTarget: 'commonjs2',
+    libraryTarget: 'umd',
   },
 
   /**
--- a/.erb/configs/webpack.config.preload.prod.babel.js
+++ b/.erb/configs/webpack.config.preload.prod.babel.js
@@ -1,0 +1,75 @@
+/**
+ * Webpack config for production electron preload process
+ */
+
+import path from 'path';
+import webpack from 'webpack';
+import { merge } from 'webpack-merge';
+import TerserPlugin from 'terser-webpack-plugin';
+import { BundleAnalyzerPlugin } from 'webpack-bundle-analyzer';
+import baseConfig from './webpack.config.base';
+import CheckNodeEnv from '../scripts/CheckNodeEnv';
+import DeleteSourceMaps from '../scripts/DeleteSourceMaps';
+
+CheckNodeEnv('production');
+DeleteSourceMaps();
+
+const devtoolsConfig = process.env.DEBUG_PROD === 'true' ? {
+  devtool: 'source-map'
+} : {};
+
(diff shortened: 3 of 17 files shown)


amilajack (commented):
Thanks for the PR! I was wondering if you could provide some more context around the motivation for the webpack target changes and the migration from `process` to `dev`.

davipk:
Sure. The webpack target was changed from ``electron-renderer`` to ``web`` because using ``electron-renderer`` uses some plugins that require node modules which aren't typically available in a browser. And since we now have set ``nodeIntegration`` to ``false`` by default (required in order for for ``contextIsolation`` to work at all), our target had to be changed to reflect that.

As for motivation behind not using ``process`` anymore, it was because exposing so also ends up exposing a lot of other variables that aren't necessarily needed for the app to function properly (ex. all the environment variables in ``process.ENV``). And since one of the reasons behind using ``preload.js`` is to limit renderers from accessing node globals, I thought it would be a good practice to only expose node-specific properties that are required at bare minimum for the renderer to work.

davipk:
Oh yeah, another thing I forgot to mention that I didn't have much time to figure out how I can configure``exposeInMainWorld`` function in ``preload.js`` in a way that other ``ts`` and ``tsx`` files can reference it's type on a particular api key that's been set. So for example, other TypeScript files should know that ``dev.isRunning`` is a boolean or undefined and ``dev.port`` is number or undefined. The only way I can think of right now is to declare a separate module with ``d.ts`` but I felt like there is a better way to approach this.

Let me know if you can think of any other ways to address this and I would love to add that feature in when I have time~

richard-dp:
> Oh yeah, another thing I forgot to mention that I didn't have much time to figure out how I can configure`exposeInMainWorld` function in `preload.js` in a way that other `ts` and `tsx` files can reference it's type on a particular api key that's been set. So for example, other TypeScript files should know that `dev.isRunning` is a boolean or undefined and `dev.port` is number or undefined. The only way I can think of right now is to declare a separate module with `d.ts` but I felt like there is a better way to approach this.
> 
> Let me know if you can think of any other ways to address this and I would love to add that feature in when I have time~

Hey, I found a fairly clean way of doing this with `html-webpack-plugin` you can have a look over here, https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/2827.

I'm really glad you made this PR as I started using/editing this template, because now I can finally get contextIsolation working correctly in the devServer.

davipk:
> 
> 
> > Oh yeah, another thing I forgot to mention that I didn't have much time to figure out how I can configure`exposeInMainWorld` function in `preload.js` in a way that other `ts` and `tsx` files can reference it's type on a particular api key that's been set. So for example, other TypeScript files should know that `dev.isRunning` is a boolean or undefined and `dev.port` is number or undefined. The only way I can think of right now is to declare a separate module with `d.ts` but I felt like there is a better way to approach this.
> > Let me know if you can think of any other ways to address this and I would love to add that feature in when I have time~
> 
> Hey, I found a fairly clean way of doing this with `html-webpack-plugin` you can have a look over here, #2827.
> 
> I'm really glad you made this PR as I started using/editing this template, because now I can finally get contextIsolation working correctly in the devServer.

oh, that looks pretty neat indeed! I will take a look and see if I can implement that into this PR when I have some time :)

src/index.html
```diff
@@ -20,13 +20,13 @@
...
 
-      if (process.env.NODE_ENV === 'development') {
+      if (dev.isRunning) {
         // Dynamically insert the DLL script in development env in the
         // renderer process
         scripts.push('../.erb/dll/renderer.dev.dll.js');
 
         // Dynamically insert the bundled app script in the renderer process
-        const port = process.env.PORT || 1212;
+        const port = dev.port ?? 1212;
```
clytras, line 29: Why change _locical or_ with _nullish coalescing operator_? What if `dev.port` is an empty string or `0` for some reason? In that case we'll end up with the port being an empty string or `0`.

davipk: Hey clytras,
I changed to nullish coalescing operator because it's behavior is a lot more predictable than ``||`` which will *short circuit* to right value when it's ``0`` or an empty string as you stated, instead of simply just checking for ``null`` or ``undefined``. Thus, I try to avoid using ``||`` and replace it with ``??`` whenever I see it because it results in less headache for long run when cases like that occur.
Though, I really don't think my change here would have altered the behavior anyway unless one really wants to have either an empty string or ``0`` as their port.

clytras: It will not alter the default behavior but in a long term it will lead unexperienced and _naive_ developers getting stuck to why things aren't working after some mistaken changes they or some other have made (_or even session environment variables lost/altered_); thus that will may lead to useless issue submissions regarding _"can't start electron"_ and related to port ending being `0` or an empty string.

I suggest we leave these defaults assingments with `||` operator. Logical/assignment `OR` operator is pretty straightforward on its behaviour and in assigmnet, that is _"continue with the next operand if current operand is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)"_, which is what we need here to eliminate the possibility of `port` ending up being `0` or en empty string.

davipk:
I have added global types as well as an example usage on how to properly use ``contextIsolation`` with ``preload.js``. Additionally, I've also included a separate ``webpack`` configuration for ``preload.js`` to reduce the file size when packaging the application. Please let me know if there are other changes that should be made in this PR. Thanks!

PS. I'm not quite sure why the test is keep failing... The error seems to be coming from react types which I'm pretty sure I did not touch.

src/index.html
davipk: Hey,
I have reverted the following change according to your feedback. Please let me know if you have any other suggestions about this PR.

Thanks~

davipk:
Closing my PR as #2866 essentially achieves the same thing as what I was trying to do.

Issue #2491, opened 2020-06-22
The compilation product is not recommended to be placed in the source directory


ganyanchuan1989:
The compilation product is not recommended to be placed in the source directory。

microCloudCode:
+1

amilajack:
Fixed in v3

PR #2827, opened 2021-04-05
bump to electron@12 and other misc updates

richard-dp:
A lot of cleanup. (Electron v12, moved transpiled code out of src dir, made it easier to change paths in webpack config, etc.)

Ran yarn upgrade. (now using Electron v12.0.2)
Edited webpack config, moving all generated code out of the src directory.
The electron application's package.json, yarn.lock, and node_modules have been moved to build/app.
The release directory has been moved to build/release.
Created src/main and src/render used by each context and moved the relevant srource files.
Added HtmlWebpackPlugin, which will inject the required js/css into index.html. This should also make it easier to use of the project with nodeIntegration set to false, with little modification, as it doesn't need to import the "process" module.
Added Clean.js to ./erb/scripts to replace "rm -rf" when running "yarn package", making this platform agnostic.
Removed DeleteSourceMaps.js as it didn't seem necessary.

PR #2866, opened 2021-05-06
bump to electron@12 and other misc updates

amilajack:
Picking up from https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/2827

Closes https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2491
Closes https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2654
Closes https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2766

A lot of cleanup. (Electron v12, moved transpiled code out of src dir, made it easier to change paths in webpack config, etc.)

- [x] Ran yarn upgrade. (now using Electron v12.0.2)
- [x] Edited webpack config, moving all generated code out of the src directory.
- [x] The electron application's package.json, yarn.lock, and node_modules have been moved to build/app.
- [x] The release directory has been moved to build/release.
- [x] Created src/main and src/render used by each context and moved the relevant srource files.
- [x] Added HtmlWebpackPlugin, which will inject the required js/css into index.html. This should also make it easier to use of the project with nodeIntegration set to false, with little modification, as it doesn't need to import the "process" module.
- [x] Added Clean.js to ./erb/scripts to replace "rm -rf" when running "yarn package", making this platform agnostic.
- [x] Removed DeleteSourceMaps.js as it didn't seem necessary.


src/main/preload.js
```diff
@@ -0,0 +1,29 @@
+const { contextBridge, ipcRenderer } = require('electron');
+
+contextBridge.exposeInMainWorld('electron', {
+  ipcRenderer: {
+    send: ipcRenderer.send,
+    sendSync: ipcRenderer.sendSync,
+    /**
+     * You are still able to use ipcRenderer in a renderer process with contextIsolation set to true.
+     * The contextBridge is what you want to use, although there is a current bug that is preventing from you calling ipcRenderer.on in a renderer process;
```
davipk, line 9: I suggest changing or removing the comment here because considering``ipcRenderer.on`` as a bug is not really an accurate statement and can be misleading as stated by the maintainer of electron on [this](https://github.com/electron/electron/issues/21437#issuecomment-573522360) discussion.

seifsg:
Looking forward to this PR. Thank you very much!

shensven:
the same expectation, thanks for your working!

WindForce225:
Looking forward to this, Thanks for the work!!

1zilc:
Looking forward to this, Thanks for the work!!

WindForce225:
Looking forward to this, Thanks for the work!!

richard-dp:
I undid this change `Removed DeleteSourceMaps.js as it didn't seem necessary.` under commit 804387bb1e16a400e2cbd72fb44f9f39501c5ec3 as it was needed, so that should be removed from the list.

Btw, why do you want to move `build/app` outside of the build directory? Imo it seems like the perfect place for it.

lucahost:
react fast-refresh is no longer auto updating after webpack compiled successfully.
you need to hard refresh (ctrl+r) to see the changes.

any possibility to get the auto-refresh behavior back @richard-dp ?

amilajack:
@richard-dp 
> I undid this change `Removed DeleteSourceMaps.js as it didn't seem necessary.` under commit [804387b](https://github.com/electron-react-boilerplate/electron-react-boilerplate/commit/804387bb1e16a400e2cbd72fb44f9f39501c5ec3) as it was needed, so that should be removed from the list.

Delete source maps was initially added to reduce pkg size but it complicates the debugging experience so it is being removed. Can you expand on why `DEBUG_PROD` requires deleting source maps? I don't think this is the case.

> Btw, why do you want to move `build/app` outside of the build directory? Imo it seems like the perfect place for it.

I initial reasoning was that `build` conventionally contains non-source assets and are usually gitignored. To avoid this confusion I was thinking the dir should be renamed to avoid the confusion or move `app/dist` to `src`.

For the time being this is okay. If there is significant confusion we can rename it later.

Issue #2963, opened 2021-08-29
Generate React Cli support for quick creation of react components


msamgan:
While creating the react application creation components, pages and layouts are among most common activities.
copy and paste the existing component and removing unwanted code is one hell of a mess.
To resolve that problem [this packahe](https://www.npmjs.com/package/generate-react-cli) is one of the finest.

PR #2964, opened 2021-08-29
added support for Generate React CLI

msamgan:
#2963

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,9 @@ Start the app in the `dev` environment:
 yarn start
 ```
 
+Electron React Boilerplate supports [Generate React CLI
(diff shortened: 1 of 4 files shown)


msamgan:
@wincent @afrojas @typesend @dustintownsend any one ??

amilajack:
I like the idea of boilerplate generation. If we are going to choose that direction we'd need an in-house solution that can meet all of our requirements for boilerplate generation. This PR doesn't seem to deliver those requirements. Specifically the generation should:

* generate a new project given a name, git url, and other metadata
* install dependencies
* possibly handle ejection

Essentially, something similar to create-react-app's generation cli. Component generation isn't in the roadmap since we're trying to keep the boilerplate as minimal as possible.

PR #2968, opened 2021-09-01
Load the `/` path, instead of `/index.html`

DhrumanGupta:
If you use react-router, having `exact` on `/` will not render anything on `/index.html`.
Removing `index.html` will still load it since it searches for `index.`[`html`,`php`] when loading, and give `/` as the path for react router.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -84,7 +84,7 @@ const createWindow = async () => {
     },
   });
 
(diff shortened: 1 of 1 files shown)


DhrumanGupta:
Uhm, sorry for this, but if its passed all the tests, is there a specific reason for not merging it?

amilajack:
This change causes the packaged version of the boilerplate to fail

PR #2976, opened 2021-09-08
Fix PORT env var not passed to webpack.config.renderer

AustEcon:
This fixes a bug whereby changing the port from 1212 -> 4343 for example leads to errors such as:
`Failed to load resource: net::ERR_CONNECTION_REFUSED    :1212/sockjs-node/info?t=1631071460786:1 `

In the electron browser console. Because it is still using port 1212.


amilajack (approved):

amilajack:
Thanks for this! Closed in https://github.com/electron-react-boilerplate/electron-react-boilerplate/commit/aa60317066507871efcd47c250b1c896d0fcbb34

PR #2992, opened 2021-09-28
feat: Change the MemoryRouter to BrowserRouter

bei-yang:
BrowserRouter makes it easier to add routes that can be rewritten using MemoryRouter.
eg: Rewriting GET /home to /index.html


amilajack:
How does it make it eaiser?

UserJHansen (changes requested):
Browser Router breaks a packaged apps, we should use HashRouter instead

PR #2996, opened 2021-10-01
Bump Electron to 13.5.1 to fix SSL certificate problem with Let's Encrypt

mqtik:
Hi, since yesterday, all apps around the world suffered a problem with Let's Encrypt; suddenly Electron couldn't recognize Let's Encrypt as a secure certificate, and led to errors like "Certificate has expired".

It started happened to me as well.

Here's the PR that the Electron team created yesterday, which's merged into 13.5.1:
https://github.com/electron/electron/pull/31216

More info of users having the problem:
https://github.com/electron/electron/issues/31212

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,7 +176,7 @@
     "css-loader": "^6.2.0",
     "css-minimizer-webpack-plugin": "^3.0.2",
     "detect-port": "^1.3.0",
-    "electron": "^13.2.1",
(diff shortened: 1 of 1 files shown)


mqtik:
Let me know if there's anything else to be done in this PR.

It should be merged as soon as possible.

amilajack:
Closed in #2999

PR #3029, opened 2021-11-04
Fix cross-env CLI commands

jfilter:
Since the migration to npm the CLI commands with `cross-env` are not working. <https://electron-react-boilerplate.js.org/docs/packaging>

For instance, `npm run cross-env DEBUG_PROD=true npm run package` fails  with:

```
npm ERR! Missing script: "cross-env"
npm ERR!
npm ERR! To see a list of scripts, run:
npm ERR!   npm run
```

 This PR adds a new script to make those commands work again.


amilajack:
The docs are broken. I need to update them￼

amilajack:
Fixed https://github.com/electron-react-boilerplate/site/commit/f733e036b91f20fb1fbf09b85ebd63c1224d5208

jfilter:
But this means `cross-env` needs to be installed globally, right?

Issue #3031, opened 2021-11-05
sqlite3 not finding the module after the start in the development side (bug in link-modules.js)


guillermoriv:
## Summary

I was checking the erb-example-sqlite, I did the same as the example but when I run the project with start this is the error that I get:

![image](https://user-images.githubusercontent.com/61250854/140506290-b0eb2077-9346-475b-90ec-ada861d345a2.png)

So it always give me that in the main process, I wanted to know if there is a work around about this ?

I have also the @types/sqlite3 into the `main-project/package.json` and the sqlite3 into the `main-project/src/release/app/package.json`, maybe if anyone can help me, thanks in advice guys!

guillermoriv:
I manage to fix the error, maybe this is a bug?, this was the issue in the `.erb` in the `scripts` folder, I was digging around and I saw the `link-modules.js` was exporting this: 
`import {
  appNodeModulesPath,
  srcNodeModulesPath,
} from '../configs/webpack.paths';`

And in reality this is going to get undefined because we are exporting a object as a default export, so the correct way to import this from this file, should be:

`import webpackPaths from '../configs/webpack.paths';`

And should access to the variables by:

`fs.symlinkSync(
    webpackPaths.appNodeModulesPath,
    webpackPaths.srcNodeModulesPath,
    'junction'
  );`

Vishal1419:
Thank you @guillermoriv 
I was facing this issue from some days and was going to move away from sqlite3 as it is mentioned in their documentation that sqlite3 is supported on electron's version 8.2.x
But you pulled me back by finding out the issue.

After following your advice, everything is working fine now. Thanks again!

guillermoriv:
Always here to help mate!

PR #3032, opened 2021-11-06
Fix: Paths when importing webpackPaths.

bugii:
This PR allows to use native modules again. Since the default export is an object, the previous version would use undefined instead of the real paths.

### DIFF
--- a/.erb/scripts/link-modules.ts
+++ b/.erb/scripts/link-modules.ts
@@ -1,8 +1,8 @@
 import fs from 'fs';
-import {
-  appNodeModulesPath,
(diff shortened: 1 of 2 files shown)


guillermoriv (approved):

guillermoriv:
[Related issue with the link-modules.js](https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/3031),  I don't know if you saw my issue and created a PR, but for history this is the issue that was created.

guillermoriv (changes requested):
For integrity maybe this need to continue to be a `.js` file and no a `.ts`

PR #3044, opened 2021-11-19
chore: drop react router

amilajack:


UserJHansen:
Why do we need to drop react router?

Isn't it useful for creating multiple page applications?

If you do decide to push changes, please remove the yarn.lock file

amilajack:
Ah good catch, thanks!

amilajack:
Closing and might revisit later

PR #3052, opened 2021-11-29
Fix type definition in webpack configs

Gu-Miao:
My VSCode shows me a lot of errors about typescript when I open webpack config files. So I try to fix them:

* I write a new interface `WebpackConfig` with `webpack.Configuration` and `webpackDevServer.Configuration` and add type assert to export config in `webpack.config.base.ts`.
* I update `webpack-dev-server` to version `4.5.0` and add `@types/webpack-dev-server@4.5.0` due to that there is no `@types/webpack-dev-server@4.4.0`.
* I remove `devServer.historyApiFallback.disableDotRule = false` in `webpack.config.renderer.dev.ts` because its default value is `false` and its type is `true | undefined`.
* By the way, the `version` property in `release/app/package-lock.json` does not match `release/app/package.json`'s, so I correct it.

I hope the problem could be resloved(not necessarily my way), after all, large sections of red wavy lines are really uncomfortable. 

Best

Gu

### DIFF
--- a/.erb/configs/webpack.config.renderer.dev.ts
+++ b/.erb/configs/webpack.config.renderer.dev.ts
@@ -170,7 +170,6 @@ export default merge(baseConfig, {
     },
     historyApiFallback: {
       verbose: true,
-      disableDotRule: false,
     },
     onBeforeSetupMiddleware() {
       console.log('Starting Main Process...');
--- a/.erb/configs/webpack.config.base.ts
+++ b/.erb/configs/webpack.config.base.ts
@@ -5,8 +5,9 @@
 import webpack from 'webpack';
 import webpackPaths from './webpack.paths';
 import { dependencies as externals } from '../../release/app/package.json';
+import WebpackConfig from './webpack.config';
 
-export default {
+export default <WebpackConfig>{
   externals: [...Object.keys(externals || {})],
 
   stats: 'errors-only',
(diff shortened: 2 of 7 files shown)


.erb/configs/webpack.config.renderer.dev.ts
```diff
@@ -170,7 +170,6 @@ export default merge(baseConfig, {
     },
     historyApiFallback: {
       verbose: true,
-      disableDotRule: false,
```
amilajack, line 173: Why change this?

amilajack (changes requested):

.erb/configs/webpack.config.renderer.dev.ts
UserJHansen: The typings state that it should be true or undefined so we should just remove it.

Typings: [typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b70b8e239fc881b20346ca659f60cfa5b08d980a/types/connect-history-api-fallback/index.d.ts#L19)

UserJHansen:
Looks good

.erb/configs/webpack.config.renderer.dev.ts
Gu-Miao: Yes, @UserJHansen is right. That's why I remove it.

amilajack:
Closed in https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/3072

PR #3066, opened 2021-12-16
Eliminate error warnings generated when configuring webpack with typescript。

1111mp:
Can refer to: [webpack](https://webpack.js.org/configuration/configuration-languages/#typescript) and [webpack-merge](https://github.com/survivejs/webpack-merge#using-with-typescript)

### DIFF
--- a/.erb/configs/webpack.config.base.ts
+++ b/.erb/configs/webpack.config.base.ts
@@ -2,7 +2,7 @@
  * Base webpack config used across other specific configs
  */
 
-import webpack from 'webpack';
+import webpack, { Configuration } from 'webpack';
 import webpackPaths from './webpack.paths';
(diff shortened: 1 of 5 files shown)


.erb/configs/webpack.config.base.ts
```diff
@@ -48,4 +48,4 @@ export default {
       NODE_ENV: 'production',
     }),
   ],
-};
+} as Configuration;
```
phuebner, line 51: I think it would be cleaner to use a typed variable instead of a type cast.

e.g.:
```ts
const configuration: webpack.Configuration = {
......
}

export default configuration;

```
I have implemented a version with this variant in pull request #3072

amilajack:
Closing in favor of https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/3072

PR #3068, opened 2021-12-17
remove react router

OleTjernstad:


UserJHansen:
Please remove the yarn lock and fix the app.tsx file so that it renders properly

PR #3123, opened 2022-01-26
IPC example from main to renderer with renderer.d.ts

paragbaxi:


UserJHansen (changes requested):
There should be an empty dependency array for the useEffect to prevent it re-running

paragbaxi:
i can't edit the commit. need to add `import { useEffect } from 'react';` to App.tsx

paragbaxi:
added main.ts sending

UserJHansen:
lgtm, can we get a workflow run?

amilajack:
Thanks for the PR. While I think this is a geat example, I'm hesitant when adding further complexity to the boilerplate example. Instead what we should do is provide an example in the docs. To get started, you can PR a md file to the [docs](https://github.com/electron-react-boilerplate/site/tree/main/docs).

PR #3130, opened 2022-01-31
fix: use setWindowOpenHandler

MokiyCodes:
on('new-window', (event,url)=>...) is deprecated in favor of setWindowOpenHandler.
This fixes a potential future bug if (and when) the new-window event is removed, by moving to setWindowOpenHandler.


UserJHansen:
lgtm

PR #3131, opened 2022-02-03
Fix typescript webpack errors

LIMPIX31:
The webpack configuration did not compile because of  `Cannot find name 'require'` and `Cannot find name '__dirname'`.
This has now been fixed.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -19,7 +19,8 @@
     "allowSyntheticDefaultImports": true,
     "resolveJsonModule": true,
     "allowJs": true,
-    "outDir": "release/app/dist"
+    "outDir": "release/app/dist",
+    "types": ["node"]
   },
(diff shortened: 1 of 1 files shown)


amilajack:
What was the error you were running into? I can't repro

LIMPIX31:
> What was the error you were running into? I can't repro

When running `yarn start` I was getting an error that `require` and `__dirname` did not exist in the webpack configuration files. I was updating my application to a new version of the ERB, I replaced all the files with new ones from the new ERB version excluding my source code.

UserJHansen:
This project does not use yarn, try using npm instead

LIMPIX31:
> This project does not use yarn, try using npm instead

I think there is no difference, because both `yarn start` and `npm start` commands run the same command.

PR #3155, opened 2022-03-05
support preload.ts

### LINKED ISSUE #3058: support preload.ts
If I rename the `preload.js` to `preload.ts` and then run `npm run start`, it will not load the preload script file.

Error:
```
Unable to load preload script: Unable to load preload script: .../src/main/preload.js
```

anniekatlin:
closes #3058


UserJHansen:
lgtm, looks like it might be able to be done better though, using more out files in the original build

anniekatlin:
> lgtm, looks like it might be able to be done better though, using more out files in the original build

Where is the original build located?

.erb/configs/webpack.config.renderer.dev.ts
```diff
@@ -162,6 +162,15 @@ const configuration: webpack.Configuration = {
       verbose: true,
     },
     onBeforeSetupMiddleware() {
+      console.log('Starting preload.js builder...');
+      spawn('npm', ['run', 'start:preload'], {
+        shell: true,
+        env: process.env,
```
amilajack, line 168: I think this may be the default. We can safely omit this.

src/main/preload.ts
```diff
@@ -1,22 +1,22 @@
-const { contextBridge, ipcRenderer } = require('electron');
+import { contextBridge, ipcRenderer } from 'electron';
 
 contextBridge.exposeInMainWorld('electron', {
   ipcRenderer: {
     myPing() {
       ipcRenderer.send('ipc-example', 'ping');
     },
-    on(channel, func) {
+    on(channel: string, func: (...args: any[]) => void) {
```
amilajack, line 8: We can suppress the linter warnings with `eslint-disable-next-line`

amilajack (changes requested):
Thanks for this PR! This has been a long awaited feature of the boilerplate. Here's a few minor comments.

anasrar:
issue that I found is webpack keep watch preload.ts after electron close in `npm start` and also nice to have `preload.d.ts` for better type checking in react side.

`src/preload.d.ts`
```ts
declare global {
  interface Window {
    electron: {
      ipcRenderer: {
        myPing(): void;
        on(channel: string, func: (...args: any[]): void;
        once(channel: string, func: (...args: any[]): void;
      };
    };
  };
};

export {};
```

amilajack:
@anasrar great feedback and good catch. @anniekatlin can you also incorporate @anasrar's feedback

anasrar:
my current solution is polling check the parent PID using webpack hook

`.erb/configs/webpack.config.preload.dev.ts`
```ts
class CheckParentPIDProcess {
  interval: number;

  constructor(interval: number) {
    this.interval = interval;
  }

  check() {
    try {
      process.kill(parseInt(process.env.PARENT_PID), 0);
    } catch (e) {
      process.exit(0);
    }
    setTimeout(() => {
      this.check();
    }, this.interval);
  }

  apply(compiler: Compiler) {
    if (!process.env.hasOwnProperty('PARENT_PID')) {
      console.log('PARENT_PID not found on environment');
      return;
    }
    compiler.hooks.entryOption.tap('CheckPIDProcess', (context, entry) => {
      console.log(`Start check parent PID process: ${process.env.PARENT_PID}`);
      // recommend using promise instead of setTimeout
      setTimeout(() => {
        this.check();
      }, this.interval);
    });
  }
}

const configuration: webpack.Configuration = {
  // ...
  plugins: [
    // ...
    new CheckParentPIDProcess(3000),
  ]
}
```

and passing `PARENT_PID` to environment in `.erb/configs/webpack.config.renderer.dev.ts` before spawn `npm run start:preload`
```ts
process.env.PARENT_PID = process.pid.toString();
```

anniekatlin:
I rebased my fork causing this PR being closed, and then reopen this PR after pushing latest commit, now it still request changing now. Please review this, thanks.

amilajack (approved):

amilajack:
@anniekatlin thanks for the contribution 🥳 this is been a long awaited feature!

shyamkumaryadav:
Thanks @anniekatlin

dddreee:
Thanks a lot!

PR #3160, opened 2022-03-10
Fix: vscode debugging

### LINKED ISSUE #428: Question: Debugging Main/Renderer Processes
how to debug main thread or node code?

i set 

``` json
"start-hot": "cross-env HOT=1 NODE_ENV=development electron --debug-brk=5858  -r babel-register -r babel-polyfill ./main.development",
```

but it does`t work ?
how can i do?

MatrikMoon:
Fixes #428

Explanation of changes:
As mentioned in #428, "Start Webpack Dev" has an out of date "endsPattern", so when used as a preLaunchTask it would prevent the next part of the launch configuration from running. It has been changed here to search for part of what is now webpack's only output.

While I am unfamiliar with yarn, I have a hunch that it offers you the option to pass arguments to scripts as was previously done in `launch.json`, but with npm that no longer seems to work. The way this pull request tackles that is to add a new `start:main:debug` script which includes the debugging arguments.

Additionally, the webpack configuration for the renderer has been changed so that it no longer automatically launches the main process as well, as this would cause the existing debug configuration to launch the main process twice, and once without the debug arguments. The start script which relied on this behavior has been changed.

This is my first-crack at fixing this issue, and I am new to this type of project, so if there are suggestions for improvements, I would love to hear and implement them. <3

### DIFF
--- a/.erb/configs/webpack.config.renderer.dev.ts
+++ b/.erb/configs/webpack.config.renderer.dev.ts
@@ -161,16 +161,6 @@ const configuration: webpack.Configuration = {
     historyApiFallback: {
       verbose: true,
     },
-    onBeforeSetupMiddleware() {
-      console.log('Starting Main Process...');
-      spawn('npm', ['run', 'start:main'], {
-        shell: true,
-        env: process.env,
-        stdio: 'inherit',
-      })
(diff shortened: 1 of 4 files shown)


MatrikMoon:
Closing in favor of #3169

PR #3168, opened 2022-03-18
Replacing ts-loader with esbuild-loader

1zilc:
ERB uses `happyPackMode:true` to improve the speed, which seems to ignore type checking. In this case, I don't know if  esbuild-loader can be used to improve compilation efficiency, looking forward to better advice


amilajack:
esbuild as a plugin very likely may increase build performance. i'm wondering if we should use swc for transpilation, similar to next.js.

amilajack:
closing in favor of further investigation and experimentation

PR #3169, opened 2022-03-20
Fix: vscode debugging (#428)

### LINKED ISSUE #428: Question: Debugging Main/Renderer Processes

nedsociety:
Fixes #428. Substitutes #3160.

## Summary
1. webpack devserver now accepts envvar `MAIN_ARGS` that is passed to main process as arguments.
2. launch.json now spawns regular `npm run start`, not `npm run start:main` and a separate webpack devserver task which caused the problem. Instead, it passes all the electron debug args via the aforementioned envvar.
3. tasks.json is removed as it's no longer needed for launch.json.

Compared to #3160 it has following advantages:
* Less intrusive changes.
* It does not break coupling between closing the electron process and closing the webpack devserver. With #3160 both `npm start` and vscode debugging causes webpack devserver to linger even if the electron process is closed which causes a major inconvenience.

## Things to consider
* Argument lexer for `MAIN_ARGS` uses regex ripped off from [SO](https://stackoverflow.com/a/19856614/3567518) and only covers the most basic syntax.
* Are there any other use cases for tasks.json besides being referred from launch.json? I'm specifically asking this because #3160 has another valid fix for tasks.json.


Langdex:
nice, it worked!
thanks

hongwen000:
Seems to work on my device.
image

amilajack (approved):

amilajack:
Thanks for this! 🥳

hongwen000:
@raphaelsetin  O(∩_∩)O Here is the list of extensions I use
- tintinweb.vscode-inline-bookmarks
- ms-vscode.powershell
- eamodio.gitlens
- ThomasKarachristos.Terminal-StatusBar
- lkytal.quicktask
- spencerwmiles.vscode-task-buttons
- SanaAjani.taskrunnercode
- maciejdems.add-to-gitignore
- ms-vscode-remote.remote-ssh-edit

raphaelsetin:
@hongwen000 Thank you so much! 😃

Issue #2141, opened 2019-01-24
Hot reload main process


kenhuang:
The current `dev` process launch `start-renderer-dev`first, then spawn a [process](https://github.com/electron-react-boilerplate/electron-react-boilerplate/blob/next/configs/webpack.config.renderer.dev.babel.js#L263) to start `start-main-dev` [process](https://github.com/electron-react-boilerplate/electron-react-boilerplate/blob/next/package.json#L30), is there a way to apply webpack config to the main process in dev mode? I think in production mode it does.

Related launch step:
```
"start-main-dev": "cross-env HOT=1 NODE_ENV=development electron -r @babel/register ./app/main.dev.js",
```

saw possiable solution from other project here:
https://github.com/input-output-hk/daedalus/blob/develop/gulpfile.js#L44

theogravity:
Confused how this isn't a thing - is it normal for devs using this project to have to kill the main process and re-build each time they want to make a change to the main process code?

ahmad-invozone:
@jeremyong Can you please describe little bit more how did you get this working locally using web-pack?

Though hot reload must be a built-in feature of this boilerplate. It's very annoying to kill the main process each time and re-build to see changes.

ahmad-invozone:
@amilajack Can we get this feature in v3.0.0, any chances?

theogravity:
I think at this point you're better off using 

https://www.electronforge.io

It's an all-in-one package like react create app that also includes hot reloading.

MrRobz:
some like this would be nice. taken from electron-forge docs
"For the main process reloading, just type `rs` in the console you launched electron-forge`
 
![image](https://user-images.githubusercontent.com/21201812/129476631-a982d57a-b6fc-44cc-b581-4507fd42afcd.png)

Jimmy-JS:
Hi @amilajack , is it working in ver 4.0.0 ? I tried to change something (global shortcut) in `src/main/main.js` but it doesn't hot reload the main process. and I also notice it's get removed from the changes list as well:

<img width="325" alt="2021-10-03 at 18 51 50@2x" src="https://user-images.githubusercontent.com/9798426/135752390-83c29409-8616-4b6b-8bb1-023f8bc2a007.png">

amilajack:
Accidentally closed this. Thanks for the heads up!

SergiiMarkin:
Hi @amilajack , 
It is very annoying to kill the main process every time I made a change in main process.
[...]

PR #3180, opened 2022-03-28
feat: hot reload main process on file change using electronmon

jgresham:
Issue: https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2141

Reasoning for introducing a new npm script command: https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2141#issuecomment-1080974540

I can add info to README if this is the route the maintainers want to go.


jgresham:
@amilajack any thoughts on this implementation?

amilajack:
We can implement this without the need for nodemon. Maybe we can merge this and drop nodemon for a simpler in house solution. I'll review soon.

package.json
```diff
@@ -10,7 +10,9 @@
     "package": "ts-node ./.erb/scripts/clean.js dist && npm run build && electron-builder build --publish never",
     "postinstall": "ts-node .erb/scripts/check-native-dep.js && electron-builder install-app-deps && cross-env NODE_ENV=development TS_NODE_TRANSPILE_ONLY=true webpack --config ./.erb/configs/webpack.config.renderer.dev.dll.ts && opencollective-postinstall",
     "start": "ts-node ./.erb/scripts/check-port-in-use.js && npm run start:renderer",
```
amilajack, line 12: Instead of introducing new scripts let's update the `start` and `start:main` scripts. This'll help keep the scripts slim.

amilajack (changes requested):

PR #3246, opened 2022-06-24
fix: Wrong img[width] units in the example app

kkomelin:
As for the [HTML documentation](https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-width), the `width` attribute of the `img` tag should only contain a non-negative integer value.

> The `width` and `height` attributes... must have values that are [valid non-negative integers](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-non-negative-integer).

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -6,7 +6,7 @@ const Hello = () => {
   return (
     <div>
       <div className="Hello">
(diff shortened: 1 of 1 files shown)


amilajack (approved):

kkomelin:
Thank you @amilajack for merging my PR.

PR #3267, opened 2022-07-18
fix clean script

Ciberusps:
clean script dont work - "clean.js" only runned in `npm run package` so its not getting any command arguments and `args` always empty

Because its only runned in `npm run package` there is no reason to leave `args` better to keep it simple

### DIFF
--- a/.erb/scripts/clean.js
+++ b/.erb/scripts/clean.js
@@ -9,7 +9,5 @@ const foldersToRemove = [
 ];
 
 foldersToRemove.forEach((folder) => {
-  if (folder !== undefined) {
-    rimraf.sync(folder);
-  }
+  rimraf.sync(folder);
 });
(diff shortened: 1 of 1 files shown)


.erb/scripts/clean.js
```diff
@@ -2,16 +2,14 @@ import rimraf from 'rimraf';
...
+  webpackPaths.buildPath,
+  webpackPaths.dllPath,
+];
 
-args.forEach((x) => {
-  const pathToRemove = commandMap[x];
-  if (pathToRemove !== undefined) {
-    rimraf.sync(pathToRemove);
+foldersToRemove.forEach((folder) => {
+  if (folder !== undefined) {
```
amilajack:
```suggestion
  if (folder) {
```

amilajack (approved):

PR #3275, opened 2022-07-31
Use semantic ts-loader options

PrinOrange:
For ts-loader, if you use `exclude` option to exclude `node_modules`, this may cause conflict between `native modules` and `project modules` during postinstall. This problem can be avoided very well by using semantic options.


amilajack:
Hey, thanks for the contribution! Can you clarify what postinstall issues are caused by the existing configuration.

PR #3283, opened 2022-08-08
Update package.json

FinsaasGH:
updated electron to 20.0.1 and electron-builder to 23.3.3


shensven:
electron@20 may cause the preload script unable to load properly
#3288

amilajack:
closing since this will likely break the preload script

PR #3284, opened 2022-08-08
Add Playwright e2e tests

mxschmitt:
Since we at Playwright got many requests in the past about electron-react-boilerplate and how to use it with Playwright, I thought it might make sense to add it to the project directly.

This runs the tests on the compiled JS with the installed Electron version, like "npx playwright main.js" would do.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,4 @@
+/** @type {import('eslint').Linter.Config} */
 module.exports = {
   extends: 'erb',
   rules: {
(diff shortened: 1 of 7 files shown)


amilajack:
Thanks for the contribution. We're trying to keep the core of ERB small. E2E tests would be best contributed as docs instead of to the core.

Would you be open to making a PR to the docs?
https://github.com/electron-react-boilerplate/docs

mxschmitt:
Sounds good! I'll try to submit a section there to the testing section inside the docs.

PR #3289, opened 2022-08-16
Skip building DLLs when required from eslint config

donovanhiland:
I noticed that DLLs were being built during the lint script and making things take longer. It's probably not a big deal usually, but when running scripts in CI the dll files don't exist before running eslint and from what I can tell they don't need to be. Figured this might be useful to put up for consideration.

closes https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/3290


amilajack (approved):
LGTM. Thanks for the contribution 🎉

PR #3291, opened 2022-08-16
chore: bump deps

amilajack:
changes

- upgrade to electron@20
- disable renderer process sandboxing
- bump all dependencies


amilajack:
passed all local testing

Issue #2522, opened 2020-07-15
Loader: Load SVG images


pranshuchittora:
How to load SVG images in React components?

nikita-skobov:
Not sure if this is the "right" way, but what I did is:

```
import repeatSvg from './repeat.svg';

<img src={repeatSvg} />
```

amilajack:
@nikita-skobov that's the right way to do this

SylarRuby:
Could we import svgs like the [create react app method](https://create-react-app.dev/docs/adding-images-fonts-and-files/#adding-svgs)? It would nice to use:

```jsx
import { ReactComponent as Logo } from './logo.svg';

function App() {
  return (
    <div>
      {/* Logo is an actual React component */}
      <Logo fill="red" />
    </div>
  );
}
```

SylarRuby:
Sorry to create nose... but I've solved it. Create React App Used SVGR:

1: yarn add @svgr/webpack

2: Inside your webpack config (.erb/configs/webpack.config.base.js)

```js
[..]
{
    test: /\.svg$/,
    use: ['@svgr/webpack'],
},
[..]
```

Then you can now:

```js
import { ReactComponent as Logo } from './logo.svg';
```

Issue #2015, opened 2018-11-19
Importing SVG as React Component


arvinsalvador:
Hello,

I am trying to import SVG as a React Component in my component. I have the following code:

```
import React, { Component } from 'react';
import Mysvg from './mysvg.svg';

class Home extends Component {
  render() {
    return (
       <Mysvg />
    );
  }
}

export default Home;
```

When I run the above code, it shows error in console:

`Warning: <data:image/svg+xml;base64,ZnVuY3Rpb24gX29iamVjdFdpdGhvdXRQcm9wZXJ0aWVzKG9iaiwga2V5cykgeyB2YXIgdGFyZ2V0ID0ge307IGZvciAodmFyIGkgaW4gb2JqKSB7IGlmIChrZXlzLmluZGV4T2YoaSkgPj0gMCkgY29udGludWU7IGlmICghT2JqZWN0LnByb3RvdHlwZS5oYXNPd25Qcm9wZXJ0eS5jYWxsKG9iaiwgaSkpIGNvbnRpbnVlOyB0YXJnZXRbaV0gPSBvYmpbaV07IH0gcmV0dXJuIHRhcmdldDsgfQoKaW1wb3J0IFJlYWN0IGZyb20gInJlYWN0IjsKZXhwb3J0IGRlZmF1bHQgKChfcmVmKSA9PiB7CiAgbGV0IHsKICAgIHN0eWxlcyA9IHt9CiAgfSA9IF9yZWYsCiAgICAgIHByb3BzID0gX29iamVjdFdpdGhvdXRQcm9wZXJ0aWVzKF9yZWYsIFsic3R5bGVzIl0pOwoKICByZXR1cm4gPHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0IiBmaWxsPSIjOGM4YThhIiB7Li4ucHJvcHN9PjxwYXRoIGZpbGw9Im5vbmUiIGQ9Ik0wIDBoMjR2MjRIMFYweiIgLz48cGF0aCBkPSJNMTkgMTNoLTZ2NmgtMnYtNkg1di0yaDZWNWgydjZoNnYyeiIgLz48L3N2Zz47Cn0pOw== /> is using incorrect casing. Use PascalCase for React components, or lowercase for HTML elements.`

[...]

PR #3307, opened 2022-09-03
add support of SVG as Component

### LINKED ISSUE #3193: Unable to load svgs as react components
## Prerequisites
Clean install of ERB - followed the advice in this issue https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2522 which led to a separate problem => 

`SvgoParserError: Non-whitespace before first tag.`

(I have validated that these svgs do not have any syntax errors)

## Expected Behavior
The following code should work `import { ReactComponent as Logout } from './assets/<any.svg>';`

## Current Behavior

```
Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

## Steps to Reproduce

Attempt to import an svg just like you would in many other react app boilerplate projects or CRA 

## Possible Solution (Not obligatory)

By default ERB should ship with the ability to load svg as react components... Still no idea how to fix this issue and it is blocking progress on my project

shyamkumaryadav:
* add `@svgr/webpack`
close #3193 #2522 #2015

### DIFF
--- a/.erb/configs/webpack.config.renderer.dev.ts
+++ b/.erb/configs/webpack.config.renderer.dev.ts
@@ -90,8 +90,27 @@ const configuration: webpack.Configuration = {
       },
       // Images
       {
-        test: /\.(png|svg|jpg|jpeg|gif)$/i,
+        test: /\.(png|jpg|jpeg|gif)$/i,
         type: 'asset/resource',
+      },
+      // SVG
+      {
+        test: /\.svg$/,
+        use: [
+          {
+            loader: '@svgr/webpack',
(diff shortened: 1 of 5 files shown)


amilajack (approved):

jgresham:
@shyamkumaryadav is there any additional work to support Jest testing components with an SVG? I think we also need this. I can open a PR with this added and add an svg import in `App` to test it out. https://react-svgr.com/docs/jest/

PR #3313, opened 2022-09-09
More resilient publish workflow options

jessedp:
This solves two challenges I faced when upgrading the publish workflow:

1. Without the signing environment vars setup in GH, the build breaks
   * I didn't test this, but I imagine defining them with incorrect values is also unhelpful
2. Building a "msi", "rpm", or "deb" under only MacOs caused build problems. Running each build on the native OS has so far allowed those to work. 

This setup (with some other stuff) is running/working over [here](https://github.com/jessedp/tablo-tools-electron/blob/main/.github/workflows/publish.ymll).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -46,5 +46,15 @@ jobs:
 
           # This is used for uploading release assets to github - GH provides it by default
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


amilajack:
Thanks for this :) 

A few questions:

> Building a "msi", "rpm", or "deb" under only MacOs caused build problems.

what build problems did you run into?

Issue #3312, opened 2022-09-09
postinstall return node-abi error after install sqlite3


Aaqu:
- [x] Using npm
- [x] Using an up-to-date [`main` branch](https://github.com/electron-react-boilerplate/electron-react-boilerplate/tree/main)
- [x] Using latest version of devtools. [Check the docs for how to update](https://electron-react-boilerplate.js.org/docs/dev-tools/)
- [x] Tried solutions mentioned in [#400](https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/400)
- [x] For issue in production release, add devtools output of `DEBUG_PROD=true npm run build && npm start`

## Expected Behavior

run `npm run postinstall` after `npm install sqlite3` in `/release/app`  work properly

## Current Behavior

return 
`Could not detect abi for version 20.0.2 and runtime electron.  Updating "node-abi" might help solve this issue if it is a new release of electron`

## Steps to Reproduce

1. `cd /release/app`
2. `npm install sqlite3`
3. `npm run postinstall`
4. `cd ../..`
5. `npm run postinstall`

## Possible Solution (Not obligatory)

package.json
```
"overrides": { 
  "electron-rebuild": { "node-abi": "3.24.0" } 
},
```
or

1. go to project folder
2. `npm uninstall electron-rebuild`
3. `npm install --save-dev electron-rebuild`
4. `cd /release/app`
5. `npm run postinstall`

## Your Environment

- Node version :  16.15.0
- electron-react-boilerplate version or branch : 4.6.0
- Operating System and version : windows 10 and 11, mac os
- Link to your project :

Aaqu:
more specifically, none-abi package in electron-rebuild needs updated to a latest version

andrewrt:
i'm seeing the same issue when installing serialport to the /releases/app folder - so it seems like the latest version breaks all native deps perhaps?

same version of serialport was working with ERB from a few months ago... so assuming it's ERB related

andrewrt:
i'll piggyback on your issue @Aaqu - here's my output: 
```
  ⨯ /Users/andrew/.nvm/versions/node/v14.17.0/bin/node process failed ERR_ELECTRON_BUILDER_CANNOT_EXECUTE
Exit code:
1
Output:

> electron-react-boilerplate@4.6.0 postinstall /Users/andrew/code/js/electron/my-cool-app/release/app
> npm run electron-rebuild && npm run link-modules


[...]

PR #3315, opened 2022-09-14
fix node-abi postinstall error

Aaqu:
#3312 
bump node-abi in electron-rebuild


Aaqu:
Now I see that I haven't created a branch. My beginnings with git and github are difficult 😔

Issue #3324, opened 2022-09-28
Why was sandbox disabled for renderer in electron v20?


aaronvg:
## Summary

Isn't the template already following guidelines on not using node integration? Sandbox still allows electron to communicate via IPC with main process no?

amilajack:
We're currently targeting commonjs. Webpack generates references to `module` which aren't supported in the sandbox env. 

Fixed in https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/3328

PR #3328, opened 2022-09-29
fix: enable sandboxing

amilajack:
Closes https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/3324


RoJoHub:
If I comment `sandbox: false`,
show`Cannot read properties of undefined (reading 'ipcRenderer')`

PR #3331, opened 2022-10-01
fix: Forbidden non-null assertion

shyamkumaryadav:
fix: Forbidden non-null assertion


amilajack (changes requested):
this changes more than just the not-null assertion. i can merge after you remove the other unrelated changes.

PR #3369, opened 2022-12-01
Fix location of `preload.js` for playwright tests

jfilter:
Without this change, the Playwright guide is not working: https://electron-react-boilerplate.js.org/docs/integration-tests

Thanks for your work!

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -75,9 +75,10 @@ const createWindow = async () => {
     height: 728,
     icon: getAssetPath('icon.png'),
     webPreferences: {
-      preload: app.isPackaged
-        ? path.join(__dirname, 'preload.js')
-        : path.join(__dirname, '../../.erb/dll/preload.js'),
+      preload:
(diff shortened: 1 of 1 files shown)


src/main/main.ts
```diff
@@ -75,9 +75,10 @@ const createWindow = async () => {
     height: 728,
     icon: getAssetPath('icon.png'),
     webPreferences: {
-      preload: app.isPackaged
```
amilajack, line 78: this will break in cases where the app is running with a prod build but hasn't been packaged

amilajack (changes requested):

src/main/main.ts
amilajack: we could try testing against `process.env.NODE_ENV === 'test'`

silvio2402:
Hey, I think this should be worked on to find a suitable solution.
Can someone create an issue?
As mentioned by @jfilter, Playwright isn't working after using the guide because of this issue.
Thanks and have a nice day!

Issue #3384, opened 2022-12-21
Use ts-config-paths-webpack-plugin


karmaral:
Hello. I found myself using this as an alternative of having to sync both the tsconfig paths and webpack's aliases. Would it be a welcome addition to implement this? There's not much change involved and so far it didn't break anything on my project.

It would save the manual labor of managing two files with the same information but different formatting.

[tsconfig-paths-webpack-plugin](https://www.npmjs.com/package/tsconfig-paths-webpack-plugin)

amilajack:
Yup, that would be a very welcome addition

PR #3387, opened 2022-12-22
Add tsconfig-paths-webpack-plugin to avoid Webpack's alias duplication

karmaral:
This PR addresses #3384.

I'm not sure about the `package-lock.json` name changing, but from a quick glance at the commit history, it seems to have changed to this auto-generated one at some point. I assume it's expected?

Also, I added a comment to hint that there's no need for using the `alias` prop now. It seemed a bit cleaner than adding a dummy example path in the `tsconfig`.

### DIFF
--- a/.erb/configs/webpack.config.base.ts
+++ b/.erb/configs/webpack.config.base.ts
@@ -3,6 +3,7 @@
  */
 
 import webpack from 'webpack';
+import TsconfigPathsPlugins from 'tsconfig-paths-webpack-plugin';
 import webpackPaths from './webpack.paths';
 import { dependencies as externals } from '../../release/app/package.json';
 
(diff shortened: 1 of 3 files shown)


tjcouch-sil:
It seems this solution does not work on the main process in development because it uses `electronmon`, not webpack. I think someone needs to figure out how to run webpack with electron hot reloading in development, but I found a workaround in the meantime:

1. Run `npm install -D tsconfig-paths`
2. Adjust the `start:main` script to require `tsconfig-paths/register` as in the following example (based on the `tsconfig-paths` readme [here](https://github.com/dividab/tsconfig-paths#with-mocha-and-ts-node)): `"start:main": "cross-env NODE_ENV=development electronmon -r ts-node/register/transpile-only -r tsconfig-paths/register ."`
3. Run `npm start`. Profit! Your aliases should now work in the main process.

See #3404 for more explanation as to what is going wrong.

PR #3412, opened 2023-02-05
chore: bump eslint-config-erb

amilajack:


Hidenaga41:
@amilajack 
Could you please tell me the reason why husky was removed?

PR #3475, opened 2023-04-30
Update README.md with example too git clone with ssh

hamzaafridi:
example for installing with ssh along with https


amilajack:
more advanced users using ssh probably already know how to clone with it instead

PR #3491, opened 2023-06-14
fix notarization: use APPLE_APP_SPECIFIC_PASSWORD instead of APPLE_ID_PASS

nsrCodes:
According to the [documentation in electron/notarize](https://github.com/electron/notarize/blob/295c39d8c0ec21efa282161c667576fc7cb39558/README.md?plain=1#L44-L62)

the `appleIdPassword` should be the app-specific-password, and **not** the password of the developer account. 

And, the [electron-builder documentation for macOS signing](https://www.electron.build/configuration/mac) mentions setting the `APPLE_APP_SPECIFIC_PASSWORD` to that value.

Screenshot 2023-06-14 at 11 25 31 PM

**Seeing as this project relies on both these tools, it would be more appropriate to use `APPLE_APP_SPECIFIC_PASSWORD` environment variable rather than `APPLE_ID_PASS` (_which is currently used_)**

### DIFF
--- a/.erb/scripts/notarize.js
+++ b/.erb/scripts/notarize.js
@@ -12,9 +12,9 @@ exports.default = async function notarizeMacos(context) {
     return;
   }
 
-  if (!('APPLE_ID' in process.env && 'APPLE_ID_PASS' in process.env)) {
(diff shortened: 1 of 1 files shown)


NicolasZanotti (approved):
@nsrCodes I have implemented the exact same changes in my own pipeline and it has been working without issue.

amilajack (approved):

PR #3492, opened 2023-06-14
feat notarization: use notarytool for code-signing

nsrCodes:
The default alt-tool has been deprecated for more than a year and will soon become unsupported 

Here's the [issue tracking the change in electron notarize](https://github.com/electron/notarize/issues/137)
And based on the [changes being made to support this](https://github.com/electron/notarize/pull/141) it will be a breaking change, that will later use `notarytool` as the default (which is also recommended by apple documentation)

This PR adds changes to now use the `notarytool` for code-signing. But this also **requires the environment to have an extra `APPLE_TEAM_ID`** variable to be set. This is the team ID you want to notarize under.

### DIFF
--- a/.erb/scripts/notarize.js
+++ b/.erb/scripts/notarize.js
@@ -22,9 +22,11 @@ exports.default = async function notarizeMacos(context) {
   const appName = context.packager.appInfo.productFilename;
 
   await notarize({
+    tool: "notarytool",
     appBundleId: build.appId,
     appPath: `${appOutDir}/${appName}.app`,
     appleId: process.env.APPLE_ID,
     appleIdPassword: process.env.APPLE_ID_PASS,
+    teamId: process.env.APPLE_TEAM_ID
   });
 };
--- a/package.json
+++ b/package.json
@@ -181,6 +181,7 @@
     ],
     "afterSign": ".erb/scripts/notarize.js",
     "mac": {
+      "notarize" : false,
       "target": {
         "target": "default",
         "arch": [


praveen-zluri:
Was looking forward to this!

jgresham:
How are `appBundleId` and `appPath` set using the tool without the script? I don't see it in the docs. It looks like Nov 1 is the deadline for updating the notary tool. https://www.electron.build/configuration/mac

.erb/scripts/notarize.js
```diff
@@ -22,9 +22,11 @@ exports.default = async function notarizeMacos(context) {
   const appName = context.packager.appInfo.productFilename;
 
   await notarize({
+    tool: "notarytool",
     appBundleId: build.appId,
     appPath: `${appOutDir}/${appName}.app`,
     appleId: process.env.APPLE_ID,
     appleIdPassword: process.env.APPLE_ID_PASS,
+    teamId: process.env.APPLE_TEAM_ID
```
NicolasZanotti, line 30: I would also check for `APPLE_TEAM_ID` and log and error in the if-statement above.

jgresham:
Using the script and setting `notarize: false` gives this   `• skipping "afterSign" hook as no signing occurred, perhaps you intended "afterPack"?`

danielduan:
in `package.json` -> `build -> mac` the `teamId` key needs to be set:
```
"notarize": {
  "teamId": "12ABC3D0FG"
}
```
the environment variables `APPLE_ID` and `APPLE_APP_SPECIFIC_PASSWORD` needs to be set

this above config in `electron-builder` will automatically notarize the build, the `notarize.js` script duplicates the notarization like @nsrCodes mentioned and can be completely removed.

amilajack:
resolved in https://github.com/electron-react-boilerplate/electron-react-boilerplate/commit/03a87c99cd6d31b2d95eb9388b8df898b13a851f

PR #3584, opened 2024-03-08
package command update. from "electron-builder build" to "electron-builder -mwl"

asadahmad27:
HI there, I wanted to contribute by updating the package command as I recently needed to make a project and package and it took me almost 2 days to make builds. it kept giving me workspace errors. then I got to know that electron-builder changed the build command to electron-builder -mwl.

So here's my code.


ddehghan:
that worked. 

The documentation on https://electron-react-boilerplate.js.org/docs/packaging is also wrong.

PR #3598, opened 2024-04-13
Update "preload.js" to "preload.ts" in main.ts

azn-abel:
Upon cloning the project and running, the preload script located at `/src/main/preload.ts` does not get executed due to a typo in this section of main.ts, which is responsible for specifying the path of the file to preload. It should be `preload.ts`, not `preload.js`.


xupea:
I don't think it will work when it's packaged since there is no .ts files at all.

PR #3619, opened 2024-06-16
fix: eslint error on new componets are fixed. new page for testing added

yusufarsln98:


yusufarsln98:
I was getting an error due to ESLINT when adding a new page. I followed the solution suggested in the issues and created a PR.

PR #3674, opened 2025-01-20
Add ELECTRON_SANDBOX param to pass --no-sandbox if needed

MaciejMalczyk:
On Ubuntu >24.04 and derivatives there's known problem with sandboxing electron apps. This is related to apparmor profile. To avoid this "--no-sandbox" is needed to run the dev/builded app. With that change you can do:
ELECTRON_SANDBOX=--no-sandbox npm start
which will start devel on ubuntu.

This of course raise some security problems but personally I don't know any other solution.

https://github.com/electron/electron/issues/18265
https://github.com/electron/electron/issues/41066

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -81,7 +81,7 @@ const createWindow = async () => {
     },
   });
 
-  //open devtools
+  // Open devtools
   if (isDebug) {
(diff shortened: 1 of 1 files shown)


MaciejMalczyk:
Forgot to create branch for change C:

PR #3680, opened 2025-02-21
bugfix: native module resolution is broken

Goli4thus:
Even though the official 'package.json' docs regarding "main" property make it seems like the recently introduced 'electronmon .' approach could work, it turns out that's not the case.

When going via 'main' property, regardless of the specified path, the actual "nodejs module resolution" will start at the '<repo>/node_modules/' level and then bubble upwards. As a result the symlinked folders '<repo>/src/node_modules' and '<repo/.erb/node_modules/' will not be considered, hence native module resolution breaks.

The solution is simply to specify the 'main.bundle.dev.js' file directly in regards to the 'electronmon' call.

Furthermore the 'main' property was removed altogether now, cause it seemingly serves no practical purpose in this template and at best is misleading if it were to stay around.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
       "url": "https://github.com/amilajack"
     }
   ],
-  "main": "./.erb/dll/main.bundle.dev.js",
   "scripts": {
     "build": "concurrently \"npm run build:main\" \"npm run build:renderer\"",
     "build:dll": "cross-env NODE_ENV=development TS_NODE_TRANSPILE_ONLY=true NODE_OPTIONS=\"-r ts-node/register --no-warnings\" webpack --config ./.erb/configs/webpack.config.renderer.dev.dll.ts",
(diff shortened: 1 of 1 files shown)


Goli4thus:
### Short status update
As of ERB template f3465ea, this change is still required to make native modules be found.


### More context
Seemingly the current status quo of the template wasn't always the case.

Pull request #3209 introduced `main` property and the `electronmon .` call approach.
But the [linked issue](https://github.com/electron/electron/issues/14076#issuecomment-447356527) re electron repo and
the resp. explanation re "package mode vs. file mode" IMO doesn't really justify this change.
The whole point there seemingly was that "file mode" happens when there's no `package.json` file present for `electron`
to pick up on and as a result some `app` object based APIs not working.
In contrast "package mode", which happens when a `package.json` file is present, is likely the desired approach.

That's all fine, but still doesn't explain why PR #3209 was merged.
I don't quite see what it improves upon. It's about `package.json` being present, not about what path `electron` (or
`electromon`, which just passes it along) is given.

But what's clear is that due to that change native module resolution broke.
Again, because that `<path>` argument passed to `electron / electronmon` seemingly very much matters to module
resolution.

Without just the `.` (dot) as the path, it starts in the current folder and looks for any `node_modules` folder at that
level. As a result the "normal `node_modules`" folder is found, but not the symlinked ones re `./src/node_modules/` and
`./.erb/node_modules`.

With this PR and the relative path to `./.erb/dll/main.bundle.dev.js`, the module resolution starts at the levels of
that bundle file. From there it "bubbles upwards" and as a result sees `./.erb/node_modules`, which contains installed native modules.

BTW: I don't know the full history of this template, but my guess is that initially the entry points was
`./src/main/main.ts`, but then later due to introducing hot-reload feature (I assume), it moved to that  `./.erb/dll/main.bundle.dev.js` based entry point.
Not quite sure what purpose the `./src/node_modules/` symlink serves atm.

> So what is the purpose of `main` property re `package.json` then?

https://www.electronjs.org/docs/latest/tutorial/tutorial-first-app#running-an-electron-app

> The main script you defined in package.json is the entry point of any Electron application. This script controls the main process, which runs in a Node.js environment and is responsible for controlling your app's lifecycle, displaying native interfaces, performing privileged operations, and managing renderer processes (more on that later).

Sure, based on those docs I see why that PR #3209 was written like it was: because that's what the docs suggest.
But even docs aren't making it clear that the actual `<path>` passed to `electron` executable still matters for module resolution.
But ERB template seemingly very intentionally passed the full relative path along in past to get symlink based native module resolution working.

BTW: I'm not saying this is obvious (not at all)!
I too needed to debug the resp. uncaught exception re native module not being found and observing a function called `
Module._resolveFilename = function(request, parent, isMain, options)` and its `let paths;` variable.
By using a conditional breakpoint ala:

> request.includes("<name_of_my_native_module>")

... I eventually saw the impact `<path>` argument had on module resolution (which then resulted in this PR here).



> Why didn't this come up earlier?

My guess is because by default no native modules are part of this template.
Furthermore, it could be that the majority of users of the template don't use native modules.

PR #3722, opened 2025-11-04
Change require to import for notarize module

gfarinacci:
Solve the following error:
Unable to `require`  moduleName=/home/... .../.erb/scripts/notarize.js message=require() of ES Module /home/... .../node_modules/@electron/notarize/lib/index.js from /... .../.erb/scripts/notarize.js not supported. That appear running:
npm run package


vv13:
Encountered the same issue; after applying the changes, it worked successfully.

amilajack (approved):

PR #3731, opened 2026-05-29
[codex] Fix Electron startup after dependency upgrade

awelm:
## Summary
- Upgrade the Electron React Boilerplate dependency set and lockfile.
- Fix startup breakages from the upgraded toolchain, including TypeScript root resolution and ESM-only chalk imports in synchronous startup scripts.
- Stop Electron devtools/extensions from auto-opening/installing during startup unless explicitly requested.
- Add the existing app icon as the renderer favicon so CDP startup checks do not report a favicon 404.

## Testing
- `npm run prestart`
- `npx prettier --check tsconfig.json .erb/configs/webpack.config.renderer.dev.ts .erb/configs/webpack.config.renderer.prod.ts .erb/scripts/check-build-exists.ts .erb/scripts/check-node-env.js .erb/scripts/check-port-in-use.js .erb/scripts/check-native-dep.js src/main/main.ts`
- `npm run build`
- `npm test -- --runInBand`
- `npm start`, then Chrome DevTools Protocol console error check at `http://localhost:1212/`: 0 errors


awelm:
@codex review

